PR #118, opened 2014-07-12
Added doc example for collections using checked:

tkhduracell:
https://github.com/bootstrap-ruby/rails-bootstrap-forms/pull/114


onebree:
When/why did this get removed? I feel the docs for this gem are lacking a bit, and removing stuff like this would explain it.

cseelus:
Yeah, had to search the issues for quite a while to stumble on this …

Otherwise great work, thanks!

PR #142, opened 2014-09-14
add role attribute for form element

spacewander:
add [role attribute](http://stackoverflow.com/questions/21922305/how-does-role-form-help-accessibility) for `form` element,
to save users from writing `:role => 'form'`.


sharshenov:
Next responses of [the same thread on Stackoverflow](https://stackoverflow.com/a/39496161) inform that
> role="form" shouldn't go on \<form\> elements

I propose to drop default value for `role` attribute https://github.com/bootstrap-ruby/bootstrap_form/pull/560

PR #183, opened 2014-12-14
Can't override primary button's classes

ttanimichi:
Hi,

Primary button classes aren't overridden when you pass classes as the option.

```ruby
= post.primary 'create', class: 'btn btn-primary disabled' # disabled class is ignored!!
```

That's becase [Hash#merge use other_hash's value](http://www.ruby-doc.org/core-2.1.5/Hash.html#method-i-merge).

```ruby
options = { class: 'btn btn-primary disabled ' }
options.merge! class: 'btn btn-primary' #=> {:class=>"btn btn-primary"}
```

[ActiveSupport's reverse_merge](http://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html#method-i-reverse_merge) takes care of this problem.

```ruby
options = { class: 'btn btn-primary disabled ' }
options.reverse_merge! class: 'btn btn-primary' #=> {:class=>"btn btn-primary disabled"}
```

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Bugfixes:
 
   - Use #underscore, not #downcase for help text scope (#140, @atipugin)
   - Radio button and checkbox labels will now include the disabled class as needed. (#156, @ScottSwezey)
+  - Can't override primary button classes (#183, @tanimichi)
 
 Features:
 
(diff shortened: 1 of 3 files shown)


potenza:
@ttanimichi Sorry for the delay! Thanks!

PR #190, opened 2015-01-06
Fix wrapper options

### LINKED ISSUE #153: wrapper class not working
I can not add class to div wrapper in a form.

veilleperso:
The wrapper options must be retrieved from css_options to work for every control and it must be merged (instead of reverse_merged) because empty form_group_options values prevent a reverse_merge to work (and if we specify things in wrapper, they should always be considered as more important than default options).

Fix #153


andrewbredow:
I'd like to second this issue! It's currently not possible to do things like set `text` and `class` for a label. I think the intention of this code is for the user-supplied options to supersede the defaults. Currently, the logic is reversed.

PR #191, opened 2015-01-16
Only wrap inputs in input-group when prepend or append is specified

bduc:
Options hash will always contain the prepend and append keys after extract! so will never be empty?

Without prepend and append the form input control has incorrect width (not 100% of container)


donv:

PR #199, opened 2015-02-13
Allow callable value_method and text_method for collection_check_boxes and collection_radio_buttons

shadwell:
Allow :value_method and :text_method for collection_check_boxes and collection_radio_buttons to be any Object that responds to , e.g. a Proc or lambda, to match the same functionality as provided by the corresponding Rails form helper methods.

Ensure that a single String checked value does not perform an include? check by converting checked value(s) to an Array.


potenza:
Awesome. Thanks @shadwell! :+1:

PR #200, opened 2015-02-13
Fix breaks on non-ActiveRecord setups

sgnn7:
Since ActiveRecord constant is undefined on configs with alternate ORMs,
we first need to check if that value is available before evaluating it.

[bootstrap-ruby/rails-bootstrap-forms#189]


potenza:
thanks @sgnn7

PR #202, opened 2015-02-19
Allow label options to be passed in as a hash

andrewbredow:
When generating fields, it should be possible to specify a hash in the
format of `label: { text: "A label", class: "a-class" }`, as indicated
in the documentation.

This approach assumes that if the developer has specified the label as a
hash, then it trumps `label_class` in the options hash.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -309,11 +309,20 @@ module BootstrapForm
       end
 
       unless options.delete(:skip_label)
-        label_class = hide_class if options.delete(:hide_label)
+        if options[:label].is_a?(Hash)
+          label_text  = options[:label].delete(:text)
+          label_class = options[:label].delete(:class)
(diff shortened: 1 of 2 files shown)


andrewbredow:
I'd love feedback on this - from a high level, the feature is just filling out functionality that's mentioned in the README (and it pretty useful). However, there is some opinion here in what the precedence of multiple options should be. Currently it's (in descending order):

1. `hide_class`
2. `label { ... }`
3. `label_text` and `label_class`
4. `label_class`

manlycode:
This is something I could really use. :+1:

trestrantham:
:+1: Is there anything I can do to help this get merged? Lost quite a bit of time debugging this until I found this fix.

mattslack:
I could really use this :+1:

laserlemon:
Looks great! :clap: @bootstrap-ruby, seems ready to :ship:. What do you think?

PR #228, opened 2015-06-05
Fix a crash when using form for Ransack

kieranklaassen:
Fix an issue that causes bootstrap form builder to crash when a field is assigned to a Ransack Search Class, by no longer assuming the object always responds to `on_validators` while figuring out whether the field is required.


pelletencate:
+1! I've been debugging this issue for hours before I found this page!

aksh0925:
+1

reubenbrown:
+1 Recently ran into this issue as well

PR #229, opened 2015-06-18
Allow to pass block to select helper in Rails >= 4.1.0

doits:
Beginning with Rails 4.1.0, the select method allows to pass a block to
render as the options string:

http://apidock.com/rails/v4.1.8/ActionView/Helpers/FormOptionsHelper/select
(the link points to 4.1.8, but it is the same for 4.1.0)

Adapt the gem to allow this.

The test is marked to run on Rails >= 4.1.0 only.

One problem remains though: When updating Rails to 4.1.0 in test
environment, many other tests fail because the order of attributes in
the output changed. I don't know how to solve this yet without rewriting
the expected output for different Rails versions (wich is a cumbersome
task ...).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Bugfixes:
 Features:
 
   - Allow primary button classes to be overridden (#183, @tanimichi)
+  - Allow to pass a block to select helper in Rails >= 4.1.0 (#229, @doits)
 
 ## 2.3.0 (2015-02-17)
 
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -66,9 +66,17 @@ module BootstrapForm
 
     alias_method_chain :file_field, :bootstrap
 
-    def select_with_bootstrap(method, choices, options = {}, html_options = {})
-      form_group_builder(method, options, html_options) do
-        select_without_bootstrap(method, choices, options, html_options)
+    if Gem::Version.new(Rails::VERSION::STRING) >= Gem::Version.new("4.1.0")
+      def select_with_bootstrap(method, choices = nil, options = {}, html_options = {}, &block)
+        form_group_builder(method, options, html_options) do
+          select_without_bootstrap(method, choices, options, html_options, &block)
+        end
+      end
+    else
+      def select_with_bootstrap(method, choices, options = {}, html_options = {})
+        form_group_builder(method, options, html_options) do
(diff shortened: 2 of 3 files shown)


mattbrictson:
:+1: 

I would also appreciate a the ability to use a block with `select`. For now I am monkey patching bootstrap_form in a similar fashion to this PR.

This feature is really useful if you want to generate custom choices for the select element, with e.g. custom `data` attributes to each choice.

lancedikson:
:+1:

lancedikson:
@manuelmeurer, check this please

manuelmeurer:
@lancedikson I'm not actively involved in development of this project.

/cc @potenza

lancedikson:
@manuelmeurer, looks like @potenza is not active on GH at this time :disappointed:

manuelmeurer:
@baldwindavid @carloslopes @sethvargo @wingrunr21 Can one of you guys have a look?

wingrunr21:
I'm also not actively involved but I don't really see anything wrong with the PR.

doits:
I'm using this for the last months and have no problems with it. Feel free to use the fork.

Only problem with this PR is that updating to Rails >= 4.1.0 in test environment makes all other tests fail, because the output order of attributes has changed and the tests depend on this. One would have to rewrite all tests to not depend on ordering, e.g. not comparing output strings exactly but either checking for attributes (`string.include?("...")`) or parsing the output (nokogiri).

Will there be a rewrite for bootstrap 4? Maybe tests will be rewritten in then, too, then this should be taken in consideration to make tests more stable.

carloslopes:
Hey guys! Sorry about the delay!

The code appears to be fine for me, but I'm a little bit concerned about what you said of the Rails 4.1.0 issue.

The other examples fail when we use 4.1.0 on test environment?

doits:
> The other examples fail when we use 4.1.0 on test environment?

Exactly!

carloslopes:
So, the lib breaks with someone uses rails 4.1.0?

doits:
No, just the tests fail due to the way they are written.

carloslopes:
Yep, I'm not currently using the gem anymore. But like different people said that this is fine, I'll merge it

carloslopes:
There's something more you need to be merged? Then I can close a new version

mattbrictson:
Hi @doits, thanks again for this PR. You mentioned how the tests fail with Rails 4.1. Do you have any suggestions for how we can fix it? I opened a new issue to discuss it here: #273

PR #235, opened 2015-08-19
Add tel_field helper

lesleh:
Adds a helper for input type="tel". Currently only works in Safari but it is a standard HTML form element tag.


mattbrictson:
Thanks for the PR! However we already support `type=tel` using `telephone_field` and its shorter alias, `phone_field`, so this change is not needed.

PR #236, opened 2015-08-31
Allow for Form Tag Helper Methods in Model-Backed Forms

vincedevendra:
Hello,

As I was using this excellent gem,  I found myself wanting to be able to mix form tag helpers into my model-backed forms.  In order to accomplish this, I check for the presence of a `tag` option, and if it's present, within `#form_group_options`, I temporarily set `@object` to `nil`, `@object_name` to an empty string, and `@acts_like_form_tag` to `true`.  The original values are passed into `#form_group` and set back to their original values before the next form helper method is called.  

Let me know what you think!

Thanks,
Vince

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Bugfixes:
 Features:
 
   - Allow primary button classes to be overridden (#183, @tanimichi)
+  - Allow users to mix form tag helper methods into model-backed forms by passing in `tag: true` to form helper methods. (@vincedevendra)
 
 ## 2.3.0 (2015-02-17)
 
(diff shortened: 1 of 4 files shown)


mattbrictson:
Hi, thanks for the PR!

Is your goal to allow things like `email_field_tag` that are Bootstrap styled?

I think that would be useful to have, but I disagree with the `tag: true` approach to accomplishing that. It still requires you use `bootstrap_form_for`, right? That contradicts how the standard Rails `*_tag` helpers work.

Is there a more natural way to approach this? For example, how to libraries like simple_form and formtastic handle it?

PR #238, opened 2015-09-17
Allow inline form groups, to group several elements in one line

jdugarte:


kierr:
This is a must have for me. I will cherry pick the commit, hoping it's merged as soon as possible.

kierr:
Seems to only work with `:horizontal` set on the main form. Not the same as `form class="form-inline"` as I expected?

jdugarte:
@kierr You're right, I hadn't tested this for non-horizontal forms. I just committed a fix for that case. Thanks for pointing it out!

kierr:
Still no luck! Can you see what the issue is?

screenshot

    <%= f.form_group label: { text: "Name" }, inline: true do %>
      <%= f.text_field :firstname, label: "First", width: "auto" %>
      <%= f.text_field :surname, label: "Last", width: "auto" %>
    <% end %>


	<div class="form-group">
	    <label class="control-label" for="staff_">Name</label>
	    <div class="form-group form-inline">
	        <div class="form-group">
	            <label class="control-label required" for="staff_firstname">First</label>
	            <input class="form-control" name="staff[firstname]" id="staff_firstname" type="text" width="auto">
	        </div>
	        <div class="form-group">
	        <label class="control-label required" for="staff_surname">Last</label>
	        <input class="form-control" name="staff[surname]" id="staff_surname" type="text" width="auto">
	    	</div>
		</div>
	</div>

jdugarte:
@kierr From the size of the image you're posting I'm guessing the issue is the width of the window you're viewing this in: bootstrap's inline forms turn to vertical layout at a width below 768px. I tried your code in a jsfiddle (http://jsfiddle.net/jdugarte/stmb6rbn). Try to resize the results frame: it changes from inline to vertical below a certain point.

This stackoverflow question (http://stackoverflow.com/questions/19592968/bootstrap-3-breakpoints-and-media-queries) might clarify more the issue of bootstrap's media queries, breakpoints and min-widths.

Does this solve the issue? If not, could you post a little bit more of the surrounding generated html, to see if I can reproduce the problem?

donv:

PR #240, opened 2015-10-03
Fixed use of concat on label and control in order to work with cells

mizalewski:
I fixed use of concat on label and control in order to work with Cells (https://github.com/apotonick/cells).
In Cells concat returns array instead of string and this code throw error about problem with convert ActiveSupport::SafeBuffer to Array.


mattbrictson:
Is this fix still needed to work with the latest version of Cells? If so I'll consider merging it in. Thanks for the PR!

mizalewski:
I tested with Cells 4.1.2 (released 2 days ago) and it works without this fix.

Issue #205, opened 2015-03-03
Remove .required class from label?


Stajor:
How can I remove .required class from label?
I want the presence validation in my main form, but i need to remove .required from search form

```
<%= f.select :status, Campaign.statuses.collect{|status, i| [t(status), i] }, {include_blank: 'All', label_class: 'control-label'}, {name: :status, wrapper_class: 'col-md-6'} %>
```

Still render with . required
```
<label class="control-label control-label required" for="campaign_status">Status</label>
```

krsyoung:
Hi @Stajor, right now it is determined solely on the field having a presence validator -- there is no way to exclude it.

A couple of thoughts:
1. we figure out how to exclude it :-)
2. you try a different version of CSS that doesn't select the Status label when in a Search form
3. avoid using the presence validator on the Status field if in Search mode

Just curious.  The Status field is required but the class is causing you grief or you don't want to show an '*'?  Do you have different logic making the Status field optional when in the Search context?

Stajor:
Yes the field is optional only in search context.
For now I can fix it with css.
I'm waiting for next update :)
Thank you for great and useful gem

Issue #143, opened 2014-09-16
evaluate validations and add required


jurrack:
wouldn't it be great if this gem would add input required attribute and label asterisk for fields with validation?

Shoroh:
If I have condition like this:
```ruby
validates :value, presence: true, if: Proc.new { |ad_property| ad_property.ad_field.required? }
```
:value is always required.

mattbrictson:
Cleaning up old unanswered issues. If this question is still relevant, please reopen.

PR #242, opened 2015-10-20
Allow manual control of label 'required' class

lucas-nelson:
Added `skip_required` as a field option to prevent the usual checks
for a PresenceValidator and the addition of a 'required' CSS class
to the label element.

This is a fix for #205 and builds on the work done in #143.

(was:)
> Added new field option `label_required` that, if present, will control
> whether the `required` CSS class is added to the label element. If not
> present, continue to use the existing logic that looks for a
> PresenceValidator on the model object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Bugfixes:
 Features:
 
   - Add support for input-group-sm and input-group-lg in prepend/append (#226, @rusanu)
+  - Added option to prevent the 'required' CSS class on labels (#205, @LucasAU)
 
 ## [2.4.0][] (2016-07-04)
 
--- a/README.md
+++ b/README.md
@@ -158,6 +158,12 @@ validator with the associated model attribute. Presently this is one of:
 ActiveRecord::Validations::PresenceValidator or
 ActiveModel::Validations::PresenceValidator.
 
+In cases where this behavior is undesirable, use the `skip_required` option:
+
+```erb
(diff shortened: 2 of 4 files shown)


krsyoung:
Hey @LucasAU!  Nice work on cleaning this up -- this feature had way more corner cases than expected.

A couple of thoughts:
* did you have any other ideas on the parameter name?  "label_required" sounds to me more like the label is required vs the parameter.  With that said, not sure I can think of an alternative name other than simply using `required` :-)
* For the documentation update, would it make more sense to use the negative case (i.e. disabling the `required` css class for a required attribute?).  The reason I say this is the same thing in the example could be accomplished with using the standard `class: 'required'` approach which might be a little cleaner.

Thanks for your work on this.

LucasAU:
Thanks for the feedback. Good point about limiting scope to the negative case.

I'll have another go at it. I think `skip_required: true` may make the most sense. It sort of falls in line with `skip_label`.

krsyoung:
All minor nits.  `skip_required: true` makes a lot of sense and per your point lines up nicely with `skip_label`.  Thanks @LucasAU!

mattbrictson:
Hi, thanks for all your work on this PR! I'd like to merge this in. Can you rebase it and fix the conflicts?

lucas-nelson:
Rebase done, conflict resolution was straight forward.

mattbrictson:
Much appreciated! ✨

PR #246, opened 2015-11-18
Add default lookup in 'help' scope for help text I18n

RST-J:
Hello,
this PR adds a default lookup for help texts in the top level `:help` scope as Rails I18n also does for `human_attribute_name`, at least in version 4.
I find this helpful for things like address components which is the particular reason why I decided to implement this and see what you think of it.

One remark: It prefers the new default lookup over the deprecated lookup using the downcased scope.


donv:

Issue #241, opened 2015-10-08
If I use bootstrap_form_tag I get fields id starts with underscore


Shoroh:
Suppose we use bootstrap_form_tag:

``` ruby
= bootstrap_form_tag url: categories_path_generator(@category), method: :get do |f|
  = f.select :region, StaticData::REGIONS.map.with_index {|index, region| [index, region] }, include_blank: true, selected: params[:region]
```

Then I got strange **id** of any field starts with underscore:

``` html
id="_region"
```

If I try to use form namespace, like this:

``` ruby
namespace: :ads_form
```

I also get strange **id** with double underscore in the middle:

``` html
id="ads_form__region"
```

I saw the code, it looks like you use form_for without model instead to use form_tag in that case. And first one makes this underscore because is waiting for model/record name.

Also I can't set id of any field manually:

``` ruby
f.text_field :region, id: 'ads_field_region'
or
f.text_field :region, html: { id: 'ads_field_region' }
```

In both cases I get id generated by form_for:

``` html
id="_region"
```

Bug or feature?

andenq:
This is probably a bug? The name also gets surrounded by brackets.

mattbrictson:
Cleaning up old unanswered issues. If this question is still relevant, please reopen.

Issue #213, opened 2015-03-30
email_field id doesn't propagate to label


typeoneerror:
If you use a custom id with email_field the label's "for" is still the default:

```
= bootstrap_form_for :purchase do |f|
  = f.email_field :email, id: "my_custom_id"
```

result:

```
<div class="form-group">
  <label class="control-label" for="purchase_email">Email</label>
  <input id="my_custom_id" class="form-control" type="email" name="purchase[email]">
</div>
```

expected:

```
<div class="form-group">
  <label class="control-label" for="my_custom_id">Email</label>
  <input id="my_custom_id" class="form-control" type="email" name="purchase[email]">
</div>
```

datWav:
I wrote fix for this bug #221

mattbrictson:
Cleaning up old unanswered issues. If this question is still relevant, please reopen.

Issue #193, opened 2015-01-27
bootstrap_form_tag and collection_select giving bad name


burnt43:
Hello, I have a form that isn't associated with an object and therefore I am using bootstrap_form_tag. On this form I have a button where I can add different search filters which will add an an additional dropdown box when clicked. This means I can have an indefinite amount of dropdowns. My code abbreviated looks like the following:

<pre>
= bootstrap_form_tag(bsf_opts(url: "/recorded_calls", method: :post)) do |f|
...
  %td= f.collection_select "durations_greater_less[]", @duration_options, :last, :first, hide_label: true
</pre>

The html that the f.collection_select generates is found and cloned and added to the page with jQuery when the user clicks a button to add a new filter. The problem is the html that is generated has what I think is an incorrect name. Below is the html that comes out. the name is "[durations_greater_less[]]". This is a problem because lets says a user makes 5 of these dropdowns, I expect to receive as params an array for durations_greater_less, but since the name is enclosed with "[]" the params is not an array, it is a single value. I expect the name to be just durations_greater_less[].

``` html
<select class=form-control name=[durations_greater_less[]] id=_durations_greater_less[]>
  <option value=gt>Greater Than</option>
  <option value=lt>Less Than</option>
</select>
```

I tried to trace this out in the code, but I got stuck in the form_group method on form_builder.rb on the line 

<pre>
control = capture(&block).to_s
</pre>

I see that capture is delegated to @template, but I do not know where @template comes from. Before  the aforementioned line I did capture(1,2) just so it would raise an exception that I had the wrong number of args. In the backtrace I saw capture was going to code in another gem, so I'm not sure if it is the other gem where the issue lies or maybe there is just some data in bootstrap-forms I need to change to get this working.

In the meantime I can hack a solution with using rails select_tag and I'll set the classes manually to have it bootstrap styled.

burnt43:
If this was too complicated, allow me to clarify with a easier to understand example. Imagine I have a form where there is a button that says "Add Fruit". When clicked it will inset a text_field named amounts[] and a dropdown a fuits named fruits[]. So you can do something like

1. Hit "Add Fruit"
2. Enter "5" in the text field
3. Select "Banana "from the drop down
4. Hit "Add Fruit"
5. Enter "2" in the text field
6. Enter "Pineapple" from the drop down
7. Hit Submit

The params will be {amounts => ["5","2"], fruits => nil}

parhs:
[...]

PR #251, opened 2016-01-22
Fix name and id for checkbox/radio

veilleperso:
When using bootstrap_form_tag, "name", "id" and "for" properties for check_box and radio_button are not properly generated, because there is no objet_name to refer to.

This patch fix that to restore proper naming of checkboxes and radio buttons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## [Pending Release][]
 
 Bugfixes:
-  - Your contribution here!
-
+  - Fix id and name generation for fields when using `bootstrap_form_tag` to use the same conventions as in Rails (Fix #266, #193, #241)
+	- In `form_group`, we allow the hash for label to include a `for` parameter
+	- In `form_group`, use the id specified for the input as the value of `for` parameter in the matching label (Fix #213)
+	
 Features:
   - The project is now tested against and compatible with the following Rails versions 4.0, 4.1, 4.2 and 5.0 (#278).
   - Your contribution here!
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -108,6 +108,16 @@ module BootstrapForm
 
     bootstrap_method_alias :time_zone_select
 
+    def hidden_field_with_bootstrap(method, options = {})
+      if acts_like_form_tag
+        options[:id] ||= method.to_s
+        options[:name] ||= method.to_s
+      end
+      hidden_field_without_bootstrap(method, options)
+    end
+
+    bootstrap_method_alias :hidden_field
+
     def check_box_with_bootstrap(name, options = {}, checked_value = "1", unchecked_value = "0", &block)
       options = options.symbolize_keys!
       options = convert_form_tag_options(name, options) if acts_like_form_tag
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -110,6 +110,7 @@ module BootstrapForm
 
     def check_box_with_bootstrap(name, options = {}, checked_value = "1", unchecked_value = "0", &block)
       options = options.symbolize_keys!
+      options = convert_form_tag_options(name, options) if acts_like_form_tag
       check_box_options = options.except(:label, :label_class, :help, :inline)
 
       html = check_box_without_bootstrap(name, check_box_options, checked_value, unchecked_value)
@@ -128,12 +129,16 @@ module BootstrapForm
       disabled_class = " disabled" if options[:disabled]
       label_class    = options[:label_class]
 
+      label_opts = { class: label_class }
+      # only add +for+ key if not nil otherwise it prevent standard behaviour of label method
+      label_opts[:for] = options[:id] unless (options[:id].nil? || options[:id].empty?)
+
       if options[:inline]
-        label_class = " #{label_class}" if label_class
-        label(label_name, html, class: "checkbox-inline#{disabled_class}#{label_class}")
+        label_opts[:class] = "checkbox-inline#{disabled_class} #{label_class}".strip
+        label(label_name, html, label_opts)
       else
         content_tag(:div, class: "checkbox#{disabled_class}") do
-          label(label_name, html, class: label_class)
+          label(label_name, html, label_opts)
         end
       end
     end
@@ -142,6 +147,10 @@ module BootstrapForm
 
     def radio_button_with_bootstrap(name, value, *args)
       options = args.extract_options!.symbolize_keys!
+      if acts_like_form_tag
+        options[:id] ||= "#{name}_#{value}"
+        options[:name] ||= name
+      end
       args << options.except(:label, :label_class, :help, :inline)
 
       html = radio_button_without_bootstrap(name, value, *args) + " " + options[:label]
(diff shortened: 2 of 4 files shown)


mattbrictson:
Thanks for the PR! It has been a while since I've touched this code, so I will need your help to be able to fully review and merge this.

Could you explain in more detail the bug you are fixing, and what changes were necessary?

Also, I'll need you to fix the test failure and add a CHANGELOG entry before this can be merged. Thanks!

veilleperso:
In current version of bootstrap_form_tag, name, id and for properties are not properly generated just because there is no object name to refer to.
Exemple from your test:
<input id="_misc" name="[misc]" type="checkbox" value="1" />
=> Here we see that something is missing in the id, ie should be objectname_misc and not just "_misc", and same for name which should be objectname[misc] instead of [misc]... I say should because it would be like that if we had an object... But when we don't have any object, then the proper way to generate all this in rails is id="misc" and name="misc" ... 
So my patch make sure that name and id are generated the rails way

The second patch allow to be able to pass "for" parameter to the label in a control group. In the current version, there is no way to specify a specific "for" parameter ... It also fix the way for parameter is generated in a control group when a specific id have been given.

I'm not fluent in english so it's hard to turn all this in proper english. I've fixed the tests

mattbrictson:
Thanks for the explanation! I'll review this week. In the meantime, can you add an entry to the CHANGELOG file as part of this PR?

veilleperso:
Matt, I've added the changelog, but I'm not sure if it's good english ;-) I also tried to reference a few tickets that seems linked to the fix in this PR. So it should Fix #266, Fix #193, Fix #241 and Fix #213

dan-klasson:
I've tried this PR and I'm still getting `name="[foo]"`

veilleperso:
Sorry to hear that Dan. I use my patch in production, and it work properly so I'm not sure why it isn't working for you. There is also a test that validate the name format. Maybe you could try to debug your code or provide an example of what you do exactly ?

veilleperso:
If you need, here is a live example:
```
    <%= bootstrap_form_tag url:'/', html: { role: 'form' } do |form| %>
      <%= form.text_field :email_id, placeholder: 'Mail ID' %>
    <% end %>
```
it converts to
```
<form role="form" action="/" accept-charset="UTF-8" method="post">
<input name="utf8" type="hidden" value="✓">
<input type="hidden" name="authenticity_token" value="eNOQVLssnDo+DWNMvsFFjx+Nuos9qXUPK857Pr2LXHmAfUqlDPPAM7R5NobLCLyJAyZCloBYaXcOf9E0IIVqBQ==">
<div class="form-group">
  <label for="email_id">Email ID</label>
  <div class="input-group"><input class="form-control" name="email_id" id="email_id" type="text"></div>
</div>
</form>
```

dan-klasson:
@veilleperso 

This is what I did:

1. cloned the repo
2. fetched the PR: `git fetch origin pull/251/muweb`
3. `git checkout muweb`
4. Added this:

      <%= bootstrap_form_tag url: listing_path(@listing) do |f| %>
          <%= f.select :foo, options_for_select([1, 2]) %>
      <% end %>

But got this output:

    <form role="form" action="/listings/1" accept-charset="UTF-8" method="post"><input name="utf8" type="hidden" value="✓">
      <div class="form-group">
        <label class="control-label" for="foo">Foo</label>
        <select class="form-control" name="[foo]" id="_foo">
          <option value="1">1</option>
          <option value="2">2</option>
        </select>
      </div>
    </form>

veilleperso:
I see: i just fixed the generation of name and id for checkbox/radio (as stated in the tile), and that extend to any input field as they work the same. But I didn't take care of select tag which use a different code. I will look into it.

dan-klasson:
Also, using the [documented](https://github.com/bootstrap-ruby/rails-bootstrap-forms#bootstrap_form_tag) way for radio buttons, makes the html come out like this:

    <input id="rent_sale_{:value=>&quot;foo&quot;}" name="rent_sale" type="radio" value="{:value=>&quot;foo&quot;}">

veilleperso:
I think that for radio_button the proper way to do it is not use :value => "foo" , but to pass the value as you would with bootstrap_form ... So you're bug seems to be much more related to the way you build your code, as you seem to have copied the exemple of an text_field to a radio_button, but they are not built the same way ... For the proper way to use radio_button, please use the same way as you do in a classic boostrap_form / rails form

dan-klasson:
@veilleperso Select works now as well. Thank you

dan-klasson:
Same problem with hidden fields though

veilleperso:
hidden_field are not processed by rails-bootstrap-form, I can't find any specific code related to hidden field, so I guess the problem is that hidden field are delegated to rails without specific processing by rails-bootstrap-form like for all other fields ... 
I don't have to time to fix that because it's not related to my patch. Once hidden_field are handled by rails-bootstrap-form, then I think that the current patch will work like for any other input field.

But feel free to address this specific issue if you need it for your own use.

mattbrictson:
@veilleperso @dan-klasson Thanks for all your help on this! Would you say this is ready for merge?

Also, if we are going to leave `hidden_field` for a future PR, could you please create a GitHub issue in the meantime that describes the problem? That way we won't lose track of it.

veilleperso:
@mattbrictson no need for a PR, i had a quick look this morning and i guess this last patch should fix the hidden_field tag... I also added a test for it.

I guess now everything is ready to merge.

dan-klasson:
Still having the problem with the hidden field. But can leave for later.

veilleperso:
Sorry Dan, but i don't see any problem on hidden field
```
    <%= bootstrap_form_tag url: '/' do |form| %>
      <%= form.hidden_field :email, value: 'hello@yahoo.com' %>
    <% end %>
```
gives me
```
<form role="form" action="/" accept-charset="UTF-8" method="post">
<input name="utf8" type="hidden" value="✓">
<input type="hidden" name="authenticity_token" value="4IAdC35B25rLlhGIb06N2ccqU9FNPnFfHyTirSu3wuDs53eNzGk7EIiqcIqt2cMJ8ftyYJixYV0BEFuW4uMUFQ==">
<input value="hello@yahoo.com" id="email" name="email" type="hidden">
</form>
```

dan-klasson:
@veilleperso hidden fields with `form_for` is not working for me. Neither is multiselect.

veilleperso:
Sorry to hear that @dan-klasson but it's outside of the scope of this patch: I only patched a few things for bootstrap_form_for only, mainly for my own needs at first... 
You now speak of problems with form_for, feel free to investigate on that subject if there are bugs there, and maybe patch them, as I didn't saw any bug on using form_for + hidden_field ... For multiselect, I don't use them, sorry.

BTW "not working" is a pretty vague concept, so it's hard to help you ;-) It would probably help if instead you show us some code that "don't work" ?

dan-klasson:
@veilleperso No worries. Not sure how to use hidden field with `form_for` and bootstrap form. `f.hidden_field` throws a nilclass exception. `hidden_field` produces `bar[]` instead of `foo[bar]`. Multiselect has the same issue as select before you fixed it.

I'll review your code changes and submit my own PR.

veilleperso:
@dan-klasson Here is an exemple of use from code in production, where "Invoice" is an activerecord object with a field customer_id
```
    <%= form_for(Invoice.first, url: '/') do |form| %>
      <%= form.hidden_field :customer_id, id: 'bill_customer_id'  %>
    <% end %>
```
The resulting html is:
```
<form class="edit_invoice" id="edit_invoice_6" action="/" accept-charset="UTF-8" method="post">
<input name="utf8" type="hidden" value="&#x2713;" />
<input type="hidden" name="_method" value="patch" /><input type="hidden" name="authenticity_token" value="4X1M4x18fTj17LUSr+j2ZGULbIDCnTGw3MUJ3e3p9HJRSyHt7WsSvVGdEOLAY+TPI9BWVGm1CgqE6XS7ZMtXPQ==" />
<input id="bill_customer_id" type="hidden" value="5" name="invoice[customer_id]" />
```

As you can see the hidden field is properly generated, using the id I manually specified, and properly building the name field

So I guess that the problem is much more in the way you write your code, than in the gem or my patch. Again, my patch didn't concern form_for, only bootstrap_form_for.

mattbrictson:
@veilleperso @dan-klasson I lost track of the discussion here. Have all the issues been resolved and is this PR good to merge? Let me know, thanks!

lcreid:
Thank you for your interest in improving `bootstrap_form`. `bootstrap_form` has been substantially changed since this PR was submitted. The upgrade to Bootstrap 4 introduced many changes that may have addressed this PR, or made it not applicable. Therefore, we're closing this PR.

If you still believe this PR is needed, please re-open it and update it to be mergeable with the current master. Our apologies for not responding sooner.

PR #254, opened 2016-01-30
Mark symbols that are IDs as error when the underlying model has an error

rbonick:
With any form field that uses an id (e.g. :role_id) with the underlying validation being on the model it references, errors are not displayed since it doesn't match the symbol on the form. 

This checks both :role and :role_id for errors and uses error messages for both.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -267,7 +267,7 @@ module BootstrapForm
     end
 
     def has_error?(name)
-      object.respond_to?(:errors) && !(name.nil? || object.errors[name].empty?)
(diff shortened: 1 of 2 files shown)


mattbrictson:
Hi, thanks for the PR and for also including a unit test!

I am hesitating to merge this is because it feels more like a workaround to me. I'm wondering if there is a more fundamental way to solve this problem.

First I want to make sure I fully understand what you are fixing here. Could you give a more detailed example of a model, validations, and form where this is a problem?

rbonick:
Hmm, the underlying issue I was having that drove this doesn't seem broken anymore (it was something to do with a collection_select tag), so I'm going to go ahead and close this out.

PR #258, opened 2016-02-22
Sanitize our <label> `for` in accordance with Rails' <input> `id`

gkop:
Fixes mismatching `<label>` `for` and `<input>` `id`.

rails-bootstrap-forms delegates generating the `<input>` `id` to Rails, but generates <label> `for` itself. Because we were not sanitizing the value we use for `<label>` `for`, we get a mismatch between the two, when, for example, the checkbox value has a space in it. This is a a11y problem because browsers do not connect a `<label>` with an inner `<input>` if the label has a `for` that doesn't match the inner `<input>`'s `id`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 Bugfixes:
 
+  - Sanitize <label> name (IE `for` attribute) in same manner that Rails sanitizes
+    the <input> `id` attribute to fix a11y issue with `for` and `id` mismatch
   - Your contribution here!
 
 Features:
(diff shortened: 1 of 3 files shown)


mattbrictson:
Excellent PR! Thanks for including a test and changelog entry.

Could you rebase this and fix the conflicts so that I can merge this in?

gkop:
Done!

mattbrictson:
Thanks! 🙇

elliotwesoff:
Yesssss thanks for this!  Super excited about this being a thing.

PR #264, opened 2016-04-15
Options for inline and label errors to remove duplicate messages and/or format them as a sentence.

PsiRadish:
Just pass a hash instead of a boolean to `inline_errors` and/or `label_errors`, containing your choice of `no_duplicates` and/or `to_sentence`.  Like so:

```
<%= bootstrap_form_for(@user, inline_errors: { no_duplicates: true, to_sentence: true }) do |f| %>
```

[Before](http://psiradish.github.io/img/rails-bootstrap-forms/before.png) and [After](http://psiradish.github.io/img/rails-bootstrap-forms/after.png) screenshot comparison.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -28,9 +28,6 @@ module BootstrapForm
       @acts_like_form_tag = options[:acts_like_form_tag]
 
       super
-      
-      p "form forin!"
-      p @inline_errors
     end
 
     FIELD_HELPERS.each do |method_name|
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -133,23 +133,19 @@ module BootstrapForm
     alias_method_chain :check_box, :bootstrap
 
     def radio_button_with_bootstrap(name, value, *args)
-      options = args.extract_options!.symbolize_keys!
-      # args << options.except(:label, :label_class, :help, :inline)
-      args << options.except(:label, :label_class, :label_title, :help, :inline) # RADISH
+      args << options.except(:label, :label_class, :label_title, :help, :inline)
       
       html = radio_button_without_bootstrap(name, value, *args) + " " + options[:label]
-      
+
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks for the PR! I appreciate the screenshots. This looks like a good feature and I would like to merge it in after I get a release out the door for Rails 5.

In the meantime, I have a few suggestions for how to improve this PR:

* Add an example and a bit of documentation of this feature to the "Validation & Errors" section of the README.
* Include some basic unit tests.

Finally, the FormBuilder class is getting really huge, and I would like to avoid making it any bigger. Is there any way you can refactor your additions so that they reside in a separate helper object, maybe?

lib/bootstrap_form/form_builder.rb
```diff
@@ -21,9 +21,9 @@ def initialize(object_name, object, template, options)
       @control_col = options[:control_col] || default_control_col
       @label_errors = options[:label_errors] || false
       @inline_errors = if options[:inline_errors].nil?
-        @label_errors != true
+        !@label_errors
```
koenpunt: `@label_errors` isn't necessarily a boolean, so I think the comparison to a boolean is valid here.

PsiRadish:
Unfortunately I haven't used Ruby or Rails since I made this PR, and even then it wasn't my strongest language/framework, so refactoring feels like a big ask.

For the same reason it might take some time (between work and Pillars of Eternity) to add unit tests, but a quick read of what's already in bootstrap_form_test.rb makes me think it's doable.

Adding documentation to the README should be no problem, excluding the aforementioned work-and-video-games time crunch.

mattbrictson:
Thanks for the reply and for your honest assessment.

If you are still interested in doing the additional work for the README and the unit tests, then I will be happy to accept the PR without the additional refactoring. (Although please also address @koenpunt's comment about `!@label_errors`.)

However, if you are no longer really committed to Ruby and Rails, I totally understand if you'd rather let this one go. No worries.

donv:
Closing since there has been no activity for 6 years.

PR #267, opened 2016-05-21
Fix alias_method_chain deprecation messages in Rails 5

### LINKED ISSUE #253: alias_method_chain depricated
When using with rails 5 on ruby 2.2.3 the following deprication warning is thrown many times:

`DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super.`

mattbrictson:
In Rails 5, `alias_method_chain` is deprecated in favor of `Module.prepend`. However, `prepend` is only available in Ruby 2.0+, and we still want to support Ruby 1.9.

Furthermore, the `*_without_bootstrap` aliases that `alias_method_chain` creates are genuinely useful for users of bootstrap_form, because there are some cases where you don't want the Bootstrap-specific markup.

Essentially, we want the traditional `alias_method_chain` behavior, but without the pesky deprecation warnings. This commit reimplements `alias_method_chain` under a new name, `bootstrap_method_alias`. This allows us to keep the behavior the same and avoid the deprecation warnings.

Fixes #253.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Bugfixes:
 
   - Minor README corrections (#184, @msmithstubbs)
+  - Fix `alias_method_chain` deprecation warnings when using Rails 5
 
 Features:
 
--- a/lib/bootstrap_form/aliasing.rb
+++ b/lib/bootstrap_form/aliasing.rb
@@ -1,0 +1,35 @@
+module BootstrapForm
+  # This module implements the old ActiveSupport alias_method_chain feature
+  # with a new name, and without the deprecation warnings. In ActiveSupport 5+,
+  # this style of patching was deprecated in favor of Module.prepend. But
+  # Module.prepend is not present in Ruby 1.9, which we would still like to
+  # support. So we continue to use of alias_method_chain, albeit with a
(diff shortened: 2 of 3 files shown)


Elektron1c97:
looks great to me :+1:

JurgenJocubeit:
This is a really well thought out fix for the deprecation notices. Well done @mattbrictson! 👍

koenpunt:
Now that Rails 5 is out, could this be released as a new version?

em77:
Until new version is out, Rails 5 users can pull directly from GitHub with this in their Gemfile:
`gem 'bootstrap_form', require: 'bootstrap_form', git: 'https://github.com/bootstrap-ruby/rails-bootstrap-forms.git'`

koenpunt:
Obviously, but I don't prefer relying on a git repo. Also, `gem 'bootstrap_form', github: 'bootstrap-ruby/rails-bootstrap-forms'` is sufficient.

mattbrictson:
> Now that Rails 5 is out, could this be released as a new version?

Done: https://rubygems.org/gems/bootstrap_form/versions/2.4.0

PR #272, opened 2016-07-03
Prevent mutating (possibly) frozen strings

danielfone:
Currently the `#form_group` method attempts to mutate the
`:control_col` option if no label is supplied. Since this option is
passed by the user, it may be a frozen string literal. Instead, we
simply avoid mutating the supplied string.

As we approach Ruby 3 and frozen string literals become more common,
this is more likely to fail.

*edit: updated changelog.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -198,10 +198,10 @@ module BootstrapForm
         control.concat(generate_icon(options[:icon])) if options[:icon]
 
         if get_group_layout(options[:layout]) == :horizontal
-          control_class = (options[:control_col] || control_col.clone)
+          control_class = options[:control_col] || control_col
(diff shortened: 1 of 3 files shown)


lib/bootstrap_form/form_builder.rb
```diff
@@ -198,10 +198,10 @@ def form_group(*args, &block)
         control.concat(generate_icon(options[:icon])) if options[:icon]
 
         if get_group_layout(options[:layout]) == :horizontal
-          control_class = (options[:control_col] || control_col.clone)
+          control_class = options[:control_col] || control_col
```
danielfone: This offers a slight memory improvement since it's only creating new string objects as needed, instead of cloning every time.

mattbrictson:
Nice improvement. Thanks! 🙇

PR #274, opened 2016-07-04
Upgrade to Rails 4.0.13 to fix datetime helpers

mattbrictson:
In Rails 4.0.0, date_select and related helpers would fail in some circumstances with `undefined method 'to_datetime' for 0:Fixnum` errors. This was causing our tests to fail. Upgrading to the latest Rails 4.0.x version fixes the bug and now all tests pass.

See: https://github.com/rails/rails/issues/25571


mattbrictson:
Merging this in so it can be part of the 2.4.0 release.

PR #269, opened 2016-06-07
Rails 5 Support

koenpunt:
With Rails 5 around the corner, I've removed deprecated functions, and made use of new functionality like `Module#prepend`.

**EDIT**

Oh well, it seems I overlooked some other PR's. (namely #267 #248)

PR #275, opened 2016-07-04
hook include of helper to loading of ActionView

koenpunt:
this prevents errors with load order of helpers when using rspec for tests.

I also think this should be the way to go for loading extensions.

This is small part of the changes from my previous PR #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Bugfixes:
   - Minor README corrections (#184, @msmithstubbs)
   - Fix `alias_method_chain` deprecation warnings when using Rails 5
   - Allow `form_group` to work with frozen string options
+  - Fix loading of ActionView helpers in combination with RSpec and `rails-controller-testing`. (see [rails-controller-testing/issues#24](https://github.com/rails/rails-controller-testing/issues/24))
 
 Features:
 
--- a/lib/bootstrap_form.rb
+++ b/lib/bootstrap_form.rb
@@ -8,4 +8,6 @@ module BootstrapForm
   end
 end
 
-ActionView::Base.send :include, BootstrapForm::Helper
+ActiveSupport.on_load(:action_view) do
+  include BootstrapForm::Helper
+end


mattbrictson:
Looks good. Do you think this warrants a mention in the changelog?

koenpunt:
I think not, because this helper is only used within ActionView, right? So without ActionView being loaded, there's no use in having these helpers available.

koenpunt:
We might consider moving it to an initializer in the railtie:

```rb
module BootstrapForm
  module Rails
    class Engine < ::Rails::Engine
      initialize "bootstrap_form.action_view" do
        ActiveSupport.on_load(:action_view) do
          include BootstrapForm::Helper
        end
      end
    end
  end
end

mattbrictson:
I am not entirely familiar with the load order problem you are addressing with this, so please go with whatever you think works best. From a code style standpoint, I prefer the `::Rails::Engine` approach, since that seems to be the "Rails way" to do it.

koenpunt:
I will try to setup a sample app which demonstrates the issue.

koenpunt:
I've setup a sample app with some instructions in the README: https://github.com/koenpunt/bootstrap-forms-test

mattbrictson:
Thanks for the sample app. I see exactly the problem you've described.

Would you like me to merge this as-is, or did you want to switch to the `::Rails::Engine` initializer?

koenpunt:
Not sure, but I had a similar issue with the SimpleForm gem, so I'll check if that's the case. If so then I will put it up on the Rails mailing list to see if can get some insight in why this happens and what the recommended way of fixing is

koenpunt:
Turns out it is a problem being introduced by `rails-controller-testing` https://github.com/rails/rails-controller-testing/issues/24#issuecomment-231738446, and using `on_load` is the recommended way of loading.
Still not sure if it should be in the initializer or not...

koenpunt:
I've got an answer from @rafaelfranca on the matter:

> It depends, if action view is not required to the gem work in a railtie. One example of this is the observers initializer we have in activeresource. You can make activeresource work without observers.

> If the gem only works if action view is present it can be in the place that do the inclusion, in the case in the lib file.

koenpunt:
Done!

mattbrictson:
Thanks! 🙇

PR #281, opened 2016-07-12
Use ActiveSupport.on_load looks to inject behavior into ActionView::Base

cbeer:
Use `ActiveSupport.on_load` was strongly suggested in https://github.com/rails/rails-controller-testing/issues/24#issuecomment-231738446.  I'm hoping it will begin resolving Rails 5 testing failures in my application.


koenpunt:
Duplicate of #275

PR #290, opened 2016-09-01
Use `File.exist?` instead of `File.exists?`

tricknotes:
To suppress the following warning from Ruby:
```
warning: File.exists? is a deprecated name, use File.exist? instead
```

`File.exist?` is defined in current supported Ruby versions.


mattbrictson:
Thanks!

PR #291, opened 2016-09-01
Fix `hide_label` option for `collection_check_boxes`

tricknotes:
When `hide_label` and `label_class` options are given to `collection_check_boxes`, `label_class` should be rendered to label of each checkbox.

This fixes an unexpected change that caused by 9a846c98.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -334,13 +334,13 @@ module BootstrapForm
       end
 
       unless options.delete(:skip_label)
+        label_class = hide_class if options.delete(:hide_label)
         if options[:label].is_a?(Hash)
-          label_text  = options[:label].delete(:text)
-          label_class = options[:label].delete(:class)
+          label_text = options[:label].delete(:text)
+          label_class ||= options[:label].delete(:class)
           options.delete(:label)
         end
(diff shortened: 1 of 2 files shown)


mattbrictson:
Thanks for the PR and for putting the time into writing a corresponding test.

However as it stands, I am not comfortable merging this PR, because the behavior seems inconsistent and counterintuitive.

> When hide_label and label_class options are given to collection_check_boxes, label_class should be rendered to label of each checkbox.

My assumption is that `:label_class` should apply to the label of the entire form group, not the label of each individual check box. And indeed, when `:hide_label` is not specified, that's how the code behaves.

But then when `:hide_label` is `true`, you want the behavior of `:label_class` to change, such that it applies to the individual check boxes instead? I don't really follow your reasoning here. Can you elaborate?

Is there a better way to accomplish your goal that has a more self-explanatory API?

lcreid:
Thank you for your interest in improving `bootstrap_form`. `bootstrap_form` has been substantially changed since this PR was submitted. The upgrade to Bootstrap 4 introduced many changes that may have addressed this PR, or made it not applicable. Therefore, we're closing this PR.

If you still believe this PR is needed, please re-open it and update it to be mergeable with the current master. Our apologies for not responding sooner.

PR #293, opened 2016-09-23
Fix offset_col to match label with for non-sm column classes

oteyatosys:
When building horizontal forms with a label_col other than col-sm-*, the offset would be calculated incorrectly. I changed the method offset_col to take other column breakers into account.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Bugfixes:
   - Allow objects without `model_name`s to act as form objects (#295, @laserlemon)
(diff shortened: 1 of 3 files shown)


mattbrictson:
Very nice! ✨ 

Can you add a CHANGELOG entry so I can merge this? Thanks!

mattbrictson:
Could you fix the merge conflict in the CHANGELOG? Please don't delete the "your contribution here" line. Thanks!

mattbrictson:
Thanks for your contribution! 🙇

PR #295, opened 2016-09-30
Support form objects that lack model names

laserlemon:
This is useful in using Bootstrap Forms with third party libraries such as Ransack, whose queries and values do not always conform fully to the Active Model API.

A class's name, while model_name is preferred, is still a good fallback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Pending Release][]
 
 Bugfixes:
+  - Allow objects without `model_name`s to act as form objects (#295, @laserlemon)
   - Your contribution here!
 
 Features:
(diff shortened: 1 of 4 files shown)


mattbrictson:
Looks good to me! Thanks for the PR and the test. 🙌 

Can you add a CHANGELOG entry so that I can merge this?

laserlemon:
@mattbrictson Thank you! I just pushed an update to the changelog.

mattbrictson:
Thanks for your contribution! 🙇

laserlemon:
Great, thank you! Do you have a scheduled time for the next release so I can check back in?

mattbrictson:
For non-critical bug fixes, I usually push a release every 6 weeks. So the next release would be early November. In the meantime you can update your Gemfile to use the master branch:

```ruby
gem "bootstrap_form", :github => "bootstrap-ruby/rails-bootstrap-forms"
```

mattbrictson:
FYI, I just released 2.5.2 with this included. 🚢

PR #301, opened 2016-10-24
Change `form_group_builder` method to public

antpaw:
I would like to have this method as a public method. With it i can pass in a block for just the input, and not care about error, help or label text. I can add docs, tests and changelog if this is considered a good idea.

Example:

``` erb
<%= f.form_group_builder :signup_ends, wrapper_class: 'js-datepicker' do %>
  <div class="row">
    <div class="col-xs-6 col-lg-4">
      <input type="text" class="form-control js-date" required="required">
    </div>
    <div class="col-xs-4 col-lg-3">
      <input type="text" class="form-control js-time" required="required">
    </div>
    <%= f.hidden_field(:signup_ends, class: 'js-timestamp')) %>
  </div>
<% end %>
```

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -216,6 +216,63 @@ module BootstrapForm
       end
     end
 
+    def form_group_builder(method, options = {}, html_options = nil)
+      options.symbolize_keys!
+      html_options.symbolize_keys! if html_options
+
+      # Add control_class; allow it to be overridden by :control_class option
+      css_options = html_options || options
+      control_classes = css_options.delete(:control_class) { control_class }
+      css_options[:class] = [control_classes, css_options[:class]].compact.join(" ")
+
+      options = convert_form_tag_options(method, options) if acts_like_form_tag
+
+      wrapper_class = css_options.delete(:wrapper_class)
+      wrapper_options = css_options.delete(:wrapper)
+      help = options.delete(:help)
+      icon = options.delete(:icon)
+      label_col = options.delete(:label_col)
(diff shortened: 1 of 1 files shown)


mattbrictson:
It has been a while since I've reviewed the code, so could you refresh my memory: what is the difference between `form_group` and `form_group_builder`? I assume that `form_group` is not what you need for your use case, but could you explain that in more detail?

I'm not opposed to making `form_group_builder` public if I first understand its intended usage. Maybe you could share a draft of the documentation you plan on adding for it? Thanks!

antpaw:
Hi, no problem, here is the difference, its much more convienient for me to use the `form_group_builder` because i don't need to care about about error, help or label elements, they will just be appended if needed

`form_group_builder`:
```html
<div class="form-group">
   <label class="control-label" for="player_is_captain">Is Captain</label>
   <div class="form-check"><label class="form-check-label" for="player_is_captain">
      <input name="player[is_captain]" type="hidden" value="0">
      <input class="form-check-input" type="checkbox" value="1" checked="checked" name="player[is_captain]" id="player_is_captain"> 
      Is Captain
   </label>
</div>
</div>
```


`form_group`:
```html
<div class="form-group">
  <div class="form-check">
     <label class="form-check-label" for="player_is_captain">
        <input name="player[is_captain]" type="hidden" value="0">
        <input class="form-check-input" type="checkbox" value="1" checked="checked" name="player[is_captain]" id="player_is_captain"> 
        Is Captain</label>
    </div>
</div>
```

mattbrictson:
Closed in favor of #289.

PR #305, opened 2016-11-10
Bootstrap v4 alpha modifications for the file input

trammel:


trammel:
Needs support for passing attributes to the span

PR #308, opened 2016-12-15
Activate html5 client side "required" validation

donv:
Set the "required" html attribute to `true` if it is unset and the corresponding attribute is required (has a presence validation in the model).

The function is disabled by default and is activated with an `html5_validations: true` option to bootstrap_form_for.

Let me know what you think :smile:

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -6,7 +6,8 @@ module BootstrapForm
     extend BootstrapForm::Aliasing
     include BootstrapForm::Helpers::Bootstrap
 
-    attr_reader :layout, :label_col, :control_col, :has_error, :inline_errors, :label_errors, :acts_like_form_tag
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks for the PR. I like the idea!

I have a few concerns:

1. Our form builder class is huge and is growing with every PR. I want to identify opportunities to refactor to avoid this project becoming spaghetti code. Is there a chance to make incremental improvements with this PR?
2. I think `html_validations` is perhaps too broad an option. Really it is just enabling the HTML5 `required` attribute (there are many other types of HTML5 validations we don't support). Should the option be renamed to be more specific?
3. Already we are adding a `required` class to the label for similar reasons. Is there a way to consolidate that code? (The structure of the code may not allow it, but I thought I would ask.)

Once we figure out these design issues, I'll also need unit tests before I can merge this.

Again, I think this would be a good feature and it is a great start. Let me know how you'd like to proceed.

GBH:
I agree that this should be a simple `required` attribute on the input (remove it from css class as you can style on attributes too). `html5_validations` seems redundant.

Issue #313, opened 2017-02-14
Allow HTML in label of field


edwardmp:
Currently, embedding HTML in the label does not work as expected (trivial example):
<img width="252" alt="schermafbeelding 2017-02-13 om 22 54 51" src="https://cloud.githubusercontent.com/assets/1686739/22917008/7c7e6b8e-f23f-11e6-9cc6-f5840021e1ca.png">

Code example:
`<%= f.text_field :name, label: "<b>Name</b>" %>`

edwardmp:
Proposed fix in PR #314

mattbrictson:
Does this work?

```erb
<%= f.text_field :name, label: "<b>Name</b>".html_safe %>
```

edwardmp:
@mattbrictson 
Yup, it seems so. Still, personally I'd rather have this kind of logic in 1 place

mattbrictson:
I understand the need to sometimes display HTML inside error messages, but we have to be careful. Blindly calling `html_safe` inside bootstrap_form is a security issue, and as such I would prefer not to go down that path.

I think perhaps Rails already has adequate tools to address this problem.

For example, for labels I think the example I gave above (using `html_safe` explicitly) is the right way to go.

For error messages, it may not be as straightforward. But using the Rails i18n system to generate the error messages, there is a way to specify HTML in interpolated messages, as explained in the Rails Guide here: http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

Error messages can be specified as follows: http://guides.rubyonrails.org/i18n.html#error-message-scopes

edwardmp:
@mattbrictson the i18n way (appending _html to the translation key doesn't work for me.

In addition, the existing codebase already calls `html_safe` (e.g. at https://github.com/bootstrap-ruby/rails-bootstrap-forms/blob/master/lib/bootstrap_form/form_builder.rb#L423) so wouldn't it be inconsequent to disallow it here but allow it there?

mattbrictson:
> In addition, the existing codebase already calls `html_safe`

That usage is slightly different: in that case the string is a concatenation of already-safe components, so there is no security issue. Whereas calling `html_safe` on a arbitrary label value is not a good idea, IMHO.

> the i18n way (appending _html to the translation key) doesn't work for me in error messages.

[...]

PR #314, opened 2017-02-14
Allow HTML in labels

edwardmp:
See issue #313.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -374,14 +374,14 @@ module BootstrapForm
 
       options[:class] = classes.compact.join(" ")
 
+      label_text = options[:text]
       if label_errors && has_error?(name)
         error_messages = get_error_messages(name)
         label_text = (options[:text] || object.class.human_attribute_name(name)).to_s.concat(" #{error_messages}")
         label(name, label_text, options.except(:text))
-      else
-        label(name, options[:text], options.except(:text))
       end
-
+      
+      label(name, label_text.try(:html_safe), options.except(:text))
     end
 
     def generate_help(name, help_text)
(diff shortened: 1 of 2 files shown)


lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -25,7 +25,7 @@ def alert_message(title, options = {})
       def error_summary
         content_tag :ul, class: 'rails-bootstrap-forms-error-summary' do
           object.errors.full_messages.each do |error|
-            concat content_tag(:li, error)
+            concat content_tag(:li, error.html_safe)
```
lcreid: I may be stretching it here (but that's what security is about):

1. Suppose a programmer wants to give good feedback to the user, and so echoes some of the user's (invalid) input back to the user in the error message.
2. A malicious user notices this and crafts invalid input that contains some sort of scripting attack

Can someone with more knowledge comment here, please (@mattbrictson)? Are error messages guaranteed to be html safe/escaped? Would it be sufficient to simply say that the errors array should be populated with html-safe strings if that's what the programmer intends?

lib/bootstrap_form/form_builder.rb
```diff
@@ -374,14 +374,14 @@ def generate_label(id, name, options, custom_label_col, group_layout)
...
       if label_errors && has_error?(name)
         error_messages = get_error_messages(name)
         label_text = (options[:text] || object.class.human_attribute_name(name)).to_s.concat(" #{error_messages}")
         label(name, label_text, options.except(:text))
-      else
-        label(name, options[:text], options.except(:text))
       end
-
+      
+      label(name, label_text.try(:html_safe), options.except(:text))
```
lcreid: Stretching it here, but if a programmer had an application where some labels were actually built from user-entered data, then a malicious user could craft a label to mount an attack. Would it be sufficient to simply say that labels that contain html have to be html safe before they're passed to the helper?

edwardmp:
@lcreid thanks for the comments. It was already discussed in #313 that there are better options to support this, so I'm going to go ahead and close this PR.

Issue #329, opened 2017-05-02
Travis CI builds fail due to Rails 5.1 release


mattbrictson:
Rails 5.1 final has been released, and our Travis CI build is automatically picking it up.

This causes problems because certain deprecated features have been completely removed in Rails 5.1. Among those are non-versioned migrations. Rails 5.1 now requires that every migration specify an explicit Rails version number.

```
Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
  class CreateUsers < ActiveRecord::Migration[4.2]
```

Failing build:

https://travis-ci.org/bootstrap-ruby/rails-bootstrap-forms/builds/227443604

mattbrictson:
@koenpunt would you be able help on this build issue? I think you probably have the most knowledge of how our Appraisal tests are set up (see #278).

koenpunt:
Sure, I can do that later today

mattbrictson:
I'll try to fix this later this week. If anyone else wants to take a crack at it in the meantime, let me know! This is fairly high priority since it is breaking our Travis builds.

koenpunt:
@mattbrictson totally forgot about this. Made a start in #330

mattbrictson:
Fixed via #330

PR #330, opened 2017-05-08
add 5.1, use more explicit version definition for existing

koenpunt:
The start of fixing #329

defining a gem as `~> 5.0` would allow bundler to install `5.1`, which is not what we want, and made the build fail all of a sudden. So I updated it to use more explicit version definitions.

Also added 5.1, which is still failing, but that will be step 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,27 +10,18 @@ gemfile:
   - gemfiles/4.0.gemfile
   - gemfiles/4.1.gemfile
   - gemfiles/4.2.gemfile
-  - gemfiles/5.0.gemfile
-  - gemfiles/5.1.gemfile
 matrix:
-  exclude:
-    - rvm: 1.9
-      gemfile: gemfiles/5.0.gemfile
-    - rvm: 2.0
-      gemfile: gemfiles/5.0.gemfile
-    - rvm: 2.1
-      gemfile: gemfiles/5.0.gemfile
-    - rvm: 1.9
-      gemfile: gemfiles/5.1.gemfile
-    - rvm: 2.0
-      gemfile: gemfiles/5.1.gemfile
-    - rvm: 2.1
-      gemfile: gemfiles/5.1.gemfile
   include:
-    # Run Danger only once
+    - rvm: 2.2.2
+      gemfile: gemfiles/5.0.gemfile
     - rvm: 2.3.1
       gemfile: gemfiles/5.0.gemfile
+      # Run Danger only once
       script: bundle exec danger
+    - rvm: 2.2.2
+      gemfile: gemfiles/5.1.gemfile
+    - rvm: 2.3.1
+      gemfile: gemfiles/5.1.gemfile
   allow_failures:
     - gemfile: gemfiles/5.1.gemfile
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ gemfile:
   - gemfiles/4.1.gemfile
   - gemfiles/4.2.gemfile
   - gemfiles/5.0.gemfile
+  - gemfiles/5.1.gemfile
 matrix:
   exclude:
     - rvm: 1.9
@@ -19,6 +20,12 @@ matrix:
       gemfile: gemfiles/5.0.gemfile
     - rvm: 2.1
       gemfile: gemfiles/5.0.gemfile
+    - rvm: 1.9
+      gemfile: gemfiles/5.1.gemfile
+    - rvm: 2.0
+      gemfile: gemfiles/5.1.gemfile
+    - rvm: 2.1
+      gemfile: gemfiles/5.1.gemfile
   include:
     # Run Danger only once
     - rvm: 2.3.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ matrix:
     - rvm: 2.3.1
       gemfile: gemfiles/5.0.gemfile
       script: bundle exec danger
+  allow_failures:
+    - gemfile: gemfiles/5.1.gemfile
 before_install:
   - gem install bundler --conservative --version '~> 1.10'
 before_script:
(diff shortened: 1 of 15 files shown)


koenpunt:
To have the build succeed for now, we could set rails 5.1 to allow failures in the build matrix

mattbrictson:
Looks good so far! Should I wait until we fix 5.1 ("step 2"), or should I merge this PR as-is?

koenpunt:
I implemented a solution for the migration version issue, but now there's an issue with minitest after the test succeeds, which fails the build:

```
t/minitest_plugin.rb:9:in `aggregated_results': wrong number of arguments (1 for 0) (ArgumentError)
```

koenpunt:
But when that's figured out I'd say let's merge, and make the failed test with rails 5.1 a separate task.

mattbrictson:
Is this a warning we need to worry about?

> Warning: Your gemfiles directive in .travis.yml is incorrect. Run `appraisal generate --travis` to get the correct configuration.

koenpunt:
> Is this a warning we need to worry about?

Where do you see that?

koenpunt:
But I think it's barking because the gemfiles are in the matrix, and not in the normal list of gemfiles.

mattbrictson:
I see it when I run `bundle exec appraisal install`.

mattbrictson:
It looks like the minitest failure is due to this bug in Rails 5.1: https://github.com/rails/rails/pull/29022

We can fix it by pinning the version of minitest to < 5.10.2, like this:

```diff
diff --git a/Appraisals b/Appraisals
index 4eb48e3..1b74b4b 100644
--- a/Appraisals
+++ b/Appraisals
@@ -23,4 +23,5 @@ end
 
 appraise "5.1" do
   gem "rails", "~> 5.1.0"
+  gem "minitest", "< 5.10.2"
 end
```

However that still leaves us with one legitimate test failure for 5.1:

```diff
Failure:
BootstrapFormTest#test_bootstrap_form_tag_allows_an_empty_name_for_checkboxes [test/bootstrap_form_test.rb:47]:
 <form accept-charset="UTF-8" action="/users" method="post" role="form">
-  <div style="margin:0;padding:0;display:inline">
-    <input name="utf8" type="hidden" value="&#x2713;"/>
-  </div>
+  <input name="utf8" type="hidden" value="&#x2713;"/>
   <div class="checkbox">
-    <label for="_misc"><input name="[misc]" type="hidden" value="0"/><input id="_misc" name="[misc]" type="checkbox" value="1"/> Misc</label>
+    <label for="misc"><input name="misc" type="hidden" value="0"/><input id="misc" name="misc" type="checkbox" value="1"/> Misc</label>
   </div>
 </form>
```

mattbrictson:
This would fix the remaining test failure, but it is a brute-force solution. I am not 100% certain that in maintains the spirit of the test.

```diff
diff --git a/test/test_helper.rb b/test/test_helper.rb
index e38085c..5900f03 100644
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -83,4 +83,9 @@ class ActionView::TestCase
       ).to_s(:color)
     }
   end
+
+  def rails_version?(spec)
+    requirement = Gem::Requirement.new(spec)
+    requirement.satisfied_by?(Gem::Version.new(Rails::VERSION::STRING))
+  end
 end

diff --git a/test/bootstrap_form_test.rb b/test/bootstrap_form_test.rb
index b2bcd5c..663aea2 100644
--- a/test/bootstrap_form_test.rb
+++ b/test/bootstrap_form_test.rb
@@ -43,7 +43,11 @@ class BootstrapFormTest < ActionView::TestCase
   end
 
   test "bootstrap_form_tag allows an empty name for checkboxes" do
-    expected = %{<form accept-charset="UTF-8" action="/users" method="post" role="form"><div style="margin:0;padding:0;display:inline"><input name="utf8" type="hidden" value="&#x2713;" /></div><div class="checkbox"><label for="_misc"><input name="[misc]" type="hidden" value="0" /><input id="_misc" name="[misc]" type="checkbox" value="1" /> Misc</label></div></form>}
+    expected = if rails_version?(">= 5.1")
+      %{<form accept-charset="UTF-8" action="/users" method="post" role="form"><input name="utf8" type="hidden" value="&#x2713;"/><div class="checkbox"><label for="misc"><input name="misc" type="hidden" value="0"/><input id="misc" name="misc" type="checkbox" value="1"/> Misc</label></div></form>}
+    else
+      %{<form accept-charset="UTF-8" action="/users" method="post" role="form"><div style="margin:0;padding:0;display:inline"><input name="utf8" type="hidden" value="&#x2713;" /></div><div class="checkbox"><label for="_misc"><input name="[misc]" type="hidden" value="0" /><input id="_misc" name="[misc]" type="checkbox" value="1" /> Misc</label></div></form>}
+    end
     assert_equivalent_xml expected, bootstrap_form_tag(url: '/users') { |f| f.check_box :misc }
   end
```

koenpunt:
A bit less brute force would be just only taking the checkbox into the conditional;

```rb
    checkbox = if ::Rails::VERSION::STRING >= '5.1'
      %{<div class="checkbox"><label for="misc"><input name="misc" type="hidden" value="0" /><input id="misc" name="misc" type="checkbox" value="1" /> Misc</label></div>}
    else
      %{<div class="checkbox"><label for="_misc"><input name="[misc]" type="hidden" value="0" /><input id="_misc" name="[misc]" type="checkbox" value="1" /> Misc</label></div>}
    end
```

Or even less with only the label.

koenpunt:
I don't think it's a problem to have it in as "brute force", because it looks like they finally made the output of a checkbox without a group or model look normal, so I don't expect other variants on this.

mattbrictson:
OK, I took a closer look at the HTML and I think the "brute force" conditional is fine. Can you take care that change?

koenpunt:
updated tests and we no longer allow failures, so should be good to go!

mattbrictson:
Excellent! Thank you 🙇

PR #331, opened 2017-05-08
let travis use container builds

koenpunt:
by disabling sudo we allow travis to run the build in a container environment, which is expected to be much quicker.


mattbrictson:
Thanks!

PR #336, opened 2017-07-05
Fix issue that would result in an error when the output buffer doesn't respond to concat().

whiskey-1:
Hello,

I did not create an issue ticket for this.  When I started looking into it, I wasn't sure where the problem was.  The edit is pretty simple.

I am using a non-standard view layer [Trailblazer Cells](http://trailblazer.to/gems/cells/) in a Rails environment.  In certain situations when using this gem, I would get 'No implicit conversion from Array into String' errors.  I traced the error back to where I made the change.  By chaining the concat() calls together, it relies on concat() to return the correct object.  By not chaining the calls together, both calls are handled the same way and the dependency is removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 ## [Pending Release][]
 
 Bugfixes:
-  - Fix an issue that occurs when the output buffer of the calling template is based on a type other than a String (i.e. an Array).
+  - Your contribution here!
 
 Features:
   - Your contribution here!
+  
+## [Pending Release][]
+  
+Bugfixes:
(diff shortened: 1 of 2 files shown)


mattbrictson:
Hi, apologies for the late response. Thanks for opening a PR to surface this issue.

Chaining `concat` together is a natural programming pattern, and we use it in several places in bootstrap_form, not just the one you identified. Without a good way to test Trailblazer Cells compatibility, I'm afraid that any fix we make now will easily be undone in future code changes.

With that in mind, I would say supporting Trailblazer Cells is probably outside the scope of this project for the time being, so I will close this for now.

PR #363, opened 2018-01-06
Increment version to 4.0.0.dev and clean up gemspec, etc.

mattbrictson:
* Increment version to 4.0.0.dev
* Clarify in gemspec that we target Bootstrap v4, no longer v3
* Clean up gemspec based on latest `bundle gem` template
* Alphabetize dependencies
* Add license
* Homepage http -> https
* Move license file to more well-known location
* Update copyright year to 2018


mattbrictson:
@lcreid, @donv, and @desheikh I know this a trivial PR, but just as a formality, could one of you give it a review? Thanks!

bootstrap_form.gemspec
```diff
@@ -9,20 +8,25 @@ Gem::Specification.new do |s|
...
+  s.homepage    = "https://github.com/bootstrap-ruby/rails-bootstrap-forms"
+  s.summary     = "Rails form builder that makes it easy to style forms using "\
+                  "Bootstrap 4"
+  s.description = "bootstrap_form is a rails form builder that makes it super "\
+                  "easy to create beautiful-looking forms using Bootstrap 4"
+  s.license     = "MIT"
 
-  s.files = Dir["{app,config,db,lib}/**/*"] + ["MIT-LICENSE", "Rakefile", "README.md"]
-  s.test_files = Dir["test/**/*"]
+  s.files = `git ls-files -z`.split("\x0").reject do |f|
```
lcreid: I like the idea of automating the generation of the file list. This looks to me like it will include in the gem some of the files that I would think of as development files, like `.gitignore` and `.travis.yml`. I don't think it's a problem, but I'm wondering  there's a reason to include these files.

mattbrictson: It does include some unnecessary stuff, but I like that it is automatic so that we are less likely to forgot to add a file in the future. It is a standard suggested by Bundler, so I feel it is safe to follow:

https://github.com/bundler/bundler/blob/fa7e4640c06d6c7d9e7c99480342f34008aa05fd/lib/bundler/templates/newgem/newgem.gemspec.tt#L31-L33

mattbrictson:
I appreciate the approval ✅ . That's exactly the process I'd like for us to follow, thanks!

PR #366, opened 2018-01-08
[v4] Cleaning up test environment for Rails 5.0+

GBH:
- [x] Removing dev dependencies from gemspec as they're not needed in main app dev environment. Rails is a proper runtime dependency now. 
- [x] Removing Appraisal and adjusting test gemfiles for TravisCI.
- [x] Updating test/dummy to Rails 5.2 (or lower, need to see what it affects)
- [x] Setting up test view inside dummy that renders bootstrap forms
- [x] ~~Removing SQLite dependency as we don't really need database for AR objects~~ (has_many needs db)
- [x] Pretty sure that nokogiri comes with Rails, should be able to remove from Gemfile 
- [x] Adjust CONTIBUTING
- [x] ~~Remove Dangerfile? Seems kinda unneeded. Rcov and Rubocop have more value tbh.~~ Not removing


mattbrictson:
Thanks for working on this! I haven't had a chance to dig into the code yet but I have a couple quick notes:

1. I would like to keep the Dangerfile around. In my experience it makes the life a maintainer easier. I don't like pestering people about CHANGELOG entries or writing tests if a bot can do it automatically.
2. We could get rid of the dummy app entirely if we want. There is a pretty straightforward way to set up Rails and an in-memory database with not much code ([see this template](https://github.com/rails/rails/blob/f17137b0a1903203b09301b8141b5baf54561ef0/guides/bug_report_templates/active_record_master.rb)), and that way we don't have to maintain a full blown app. Seems like a win, but I'd like to hear your thoughts. I'll open a PR (I have the code ready to go) if you agree that is a good path forward.

.travis.yml
```diff
@@ -1,28 +1,13 @@
-sudo: false
 language: ruby
 rvm:
-  - 1.9
-  - 2.0
-  - 2.1
-  - 2.2
-  - 2.3.1
+  - 2.2.7
```
mattbrictson: Can we drop 2.2 as well?

.travis.yml
```diff
@@ -1,28 +1,13 @@
...
+  - 2.3.5
+  - 2.4.2
+  - 2.5.0
```
mattbrictson: What do you think about adding `ruby-head`?

.travis.yml
GBH: why? Rails 5.0+ and 5.2+ is still 2.2.2+ We're targeting that: https://rubygems.org/gems/rails/

.travis.yml
GBH: Only if you're adding Rails.master. You're testing a dependency of a dependency at that point. Plus you need to allow failures of it too. Not much of a value.

GBH:
@mattbrictson I'll put danger thing back into travis config. Dummy app is valuable for debugging html generated by the gem as you can see output in directly the browser. Way easier to work with. I got dummy working, so I'll add basic form there.

.travis.yml
mattbrictson: Ruby 2.2 will [officially be EOL at the end of March 2018](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/), and so I saw this as an opportunity to prune. I would like to be in the habit of aggressively dropping support for EOL Rubies. Even if Rails supports it, I'd be fine dropping it now, a couple months ahead of schedule.

.travis.yml
mattbrictson: I don't like getting blindsided by bugs whenever a new version of Ruby is released (even a prominent project like spring [is susceptible](https://github.com/rails/spring/issues/550), it seems), and therefore would appreciate the advance notice that `ruby-head` would give us. And indeed, Rails `5-1-stable` does in fact test against `ruby-head`. But it is an allowed failure. If we that means we have to make it an allowed failure as well, than I agree it is of little value. I'll leave it up to you. Thanks for the feedback!

mattbrictson:
> Dummy app is valuable for debugging html generated by the gem as you can see output in directly the browser.

I agree this is valuable, so long as you don't think the dummy app will be a maintenance burden. Knowing that maintainers of this project come and go, it is something to consider.

If we keep it around, could you make sure instructions for making using of the dummy app as an aid to development is explained in `CONTRIBUTING.md`? That would make it even more valuable. Thanks!

.travis.yml
GBH: If Rails still supports 2.2.2 I don't see why we should restrict it. Maybe some still choose to run 2.2 for some crazy reason. When Rails drops it, we can drop it too.

.travis.yml
GBH: Easy enough to add ruby-head. Travis will run a bit slower. I'll add one bleeding-edge job.

GBH:
@mattbrictson Neah, dummy app is generally not a big deal. Sometimes you need to tweak application config to work for all Rails versions. I have it running for 5.0, 5.1, 5.2 with very few tweaks. I left README in there explaining what files were changed. Contributing.md is edited to include bit about how to use dummy app for dev/debugging. 

I think this PR is ready to be reviewed.

CONTRIBUTING.md
```diff
@@ -1,53 +1,53 @@
...
-please read [issue #361](https://github.com/bootstrap-ruby/rails-bootstrap-forms/issues/361).
-Your comments are welcome.
-
-Thanks so much for considering contributing to bootstrap_form!
-We love pull requests!
+The release of Bootstrap 4 and Rails 5.1 have implications for the future
+direction of `bootstrap_form`. Don't worry. We plan to move this gem forward to
+Bootstrap 4 and to support Rails 5.1 and beyond. If you're thinking of
+contributing to `bootstrap_form`, please read
+[issue #361](https://github.com/bootstrap-ruby/rails-bootstrap-forms/issues/361).
```
mattbrictson: I'd like to keep the friendly and welcoming attitude (perhaps with some grammar edits):

```
Your comments are welcome.

Thanks so much for considering a contribution to bootstrap_form.
We love pull requests!
```

Any reason to remove it?

Gemfile
```diff
@@ -1,17 +1,18 @@
...
-gem "jquery-rails"
+# Uncomment and change rails version for testing purposes
+# gem "rails", "~> 5.2.0.beta2"
 
-# Declare any dependencies that are still in development here instead of in
-# your gemspec. These might include edge Rails or gems from your path or
-# Git. Remember to move these dependencies to your gemspec before releasing
-# your gem to rubygems.org.
+group :development do
+  gem "htmlbeautifier"
```
mattbrictson: Nice addition ✨

Gemfile
```diff
@@ -1,17 +1,18 @@
...
+end
 
-# To use debugger
-# gem 'ruby-debug19', :require => 'ruby-debug'
+group :test do
+  gem "diffy"
+  gem "equivalent-xml"
+  gem "mocha"
+  gem "sqlite3"
+  gem "timecop", "~> 0.7.1"
```
mattbrictson: Could you explain further why these dependencies belong in the Gemfile and not the as dev dependencies in the gemspec? It means that we have to repeat these in each of the test gemfiles, and I don't see much benefit.

.travis.yml
lcreid: I'm for supporting what Rails supports, back to Rails 5.0 as we've agreed.

.travis.yml
```diff
@@ -1,28 +1,30 @@
...
 language: ruby
 rvm:
-  - 1.9
-  - 2.0
-  - 2.1
-  - 2.2
-  - 2.3.1
+  - 2.2.7
+  - 2.3.5
+  - 2.4.2
```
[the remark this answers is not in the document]
lcreid: Supporting the latest patch release makes sense to me.

bootstrap_form.gemspec
```diff
@@ -1,8 +1,7 @@
-lib = File.expand_path("../lib", __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+$LOAD_PATH.unshift File.expand_path("../lib", __FILE__)
```
lcreid: I could be wrong, but this and a few other lines in the file seem to be reverting some changes that @mattbrictson did a couple of days ago with PR #363. Is there a reason why one way is better than the other, or am I misreading this?

CONTRIBUTING.md
lcreid: I think we want to make the CONTRIBUTING document more complete, rather than less.

Gemfile
GBH: This is mostly to keep gemspec clean. Unless you're directly working on the gem you don't need to install or even care about those dev dependencies. Gives you flexibility to cram whatever you want into your own Gemfile. See what Rails does: https://github.com/rails/rails/blob/master/rails.gemspec Million of gems used for development and testing are defined in Gemfile instead.

CONTRIBUTING.md
GBH: Felt like it was kinda redundant. I'll add it back.

.travis.yml
```diff
@@ -1,28 +1,30 @@
...
 gemfile:
-  - gemfiles/4.0.gemfile
-  - gemfiles/4.1.gemfile
-  - gemfiles/4.2.gemfile
+  - test/gemfiles/5.0.gemfile
+  - test/gemfiles/5.1.gemfile
+  - test/gemfiles/5.2.gemfile
+before_install:
+  - gem update --system
+  - gem update bundle
```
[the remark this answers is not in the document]
GBH: It updates rubygems. Needed it to get ruby 2.5 running.

bootstrap_form.gemspec
GBH: Brand new freshly generated gem will have `$:.push File.expand_path("lib", __dir__)`. Same thing really. I can't think of a reason why there needs to be a check for existence.

.travis.yml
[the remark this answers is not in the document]
GBH: didn't check the latest. will change.

Gemfile
mattbrictson: Interesting, I had no idea that was a common practice! I am still confused because the definition of gemspec dev dependencies is that they ["aren't installed by default and aren't activated when a gem is required"](https://github.com/rubygems/guides/blob/543ae72b78678d4f9b978f7070635d553687ec38/specification-reference.md#add_development_dependency). In other words, they are not considered at all unless you are working directly on the gem, which is exactly the behavior we want. Other popular Ruby projects like [capybara](https://github.com/teamcapybara/capybara/blob/e0afe31a1e4881b1a3cf082caba1fb43f311592e/capybara.gemspec#L30-L41), [sprockets](https://github.com/rails/sprockets/blob/9d945fbe1342136edd80888eba1c22f3d99d4ff9/sprockets.gemspec#L17-L33), and [rack](https://github.com/rack/rack/blob/2e92a25342b5d51f4094366c3b8dd797cbd208a4/rack.gemspec#L31-L33) do it the gemspec way, so maybe it is just personal preference?

mattbrictson: BTW I don't mean to be argumentative or wish to hold up this PR in any way. I am honestly just really curious why the Ruby community seems to have two different ways of accomplishing the same thing. Maybe the Gemfile is preferable when you have to platform-specific dependencies, and in that case the Gemfile DSL is more expressive? Feel free to ignore my ramblings. 😀

bootstrap_form.gemspec
```diff
@@ -1,8 +1,7 @@
-lib = File.expand_path("../lib", __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+$LOAD_PATH.push File.expand_path("lib", __dir__)
```
mattbrictson: Dependency declarations aside (if you want to move them to the Gemfile, no problem), and with the exception of the addition of the `required_ruby_version` (nice catch, thanks), I would prefer the other gemspec changes be reverted unless you have a strong objection.

My overall reasoning is that the existing gemspec is based on the template provided by `bundle gem`, which presumably has many eyes on it has been well vetted by the community. It has also served me well on many gems I work on.

More specifically, according to the [official docs](https://github.com/rubygems/guides/blob/543ae72b78678d4f9b978f7070635d553687ec38/specification-reference.md):

* `platform` should be left as the default value "unless the code will only compile on a certain type of system"
* `files` should not include tests (not a strict requirement, but it is common practice to exclude them)
* `bindir = "exe"` is a good safety measure to ensure binstubs or other helper scripts that we may add to `bin/` in the future are not inadvertently packaged with the gem

.travis.yml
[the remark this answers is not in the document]
mattbrictson: Yep, I totally understand the need for `gem update --system` to get 2.5.0 working. I ran into that myself. I was referring to the `bundle` line. I'd prefer:

```yaml
before_install:
  - gem update --system
  - gem install bundler --no-document
```

bootstrap_form.gemspec
GBH: Ah neat. I just used whatever `rails plugin new` would produce with some tweaks. I reverted gemspec to more or less original state minus dev dependencies. I still don't undertand load_path checking is for, but it doesn't hurt I suppose.

GBH: Gotcha on `gem install bundler --no-document`

test/dummy/.ruby-version
```diff
@@ -0,0 +1 @@
+2.4.1
```
desheikh: Should we not just use 2.5.0?
Thanks for the contribution! This is great work!

GBH: I think generator crammed it there. Removed it as you should be able to use whatever ruby you want.

GBH:
This PR is ready to be merged, unless there are some other comments?

mattbrictson:
Thanks for the PR and thanks for the reviews! It sounds like this is good to merge. Onward to 4.0!

PR #278, opened 2016-07-05
Testing multiple versions of Rails

koenpunt:
This fixes #273

PR #357, opened 2017-12-05
Use option[:id] for label's *for* attribute

duleorlovic:
If provided, use html option `id` to specify `for` attribute on label
for (text, email...) fields, checkbox, radio and select.
Solves #213, #221, #342, #349 issues and supercede #221 and #343 pull
requests.

Partially solves #251 (only the part regarding the `id`.

PR #369, opened 2018-01-09
Support for Rails 5.1 `form_with`

### LINKED ISSUE #326: Support Rails 5.1+ `form_with` style (`form_for` will eventually go away in Rails 6.x)
https://github.com/rails/rails/issues/25197

lcreid:
This PR adds a new helper bootstrap_form_with to support the Rails 5.1 form_with helper (#326). 

This PR introduces a new helper `bootstrap_form_with` that functions in a very similar way to `bootstrap_form_for` and `bootstrap_form_tag`. The user can choose to use whichever helper for a given file, although they must be using Rails 5.1 to use `bootstrap_form_with`, because it calls Rails' `form_with`. Rails 5.0 is still supported.

I would like to continue to improve the test coverage, but I believe it's good enough for people to start experimenting with.

The code changes to the actual helpers are not extensive, but testing is a challenge. `form_with` adds to the variety of expected outputs that our tests have to handle. Some examples:

* Prior to Rails 5, tabs were output in alphabetical order. After, they weren't
* Some helpers starting with Rails 5 produced more output, e.g. an empty hidden input on file tags (The genesis of assert_equivalent_xml is found in #273 and #278, which specifically addressed these first two issues.)
* Rails 5.1. stopped generating default DOM ids in the form helpers, and stopped putting the new_class HTML class on forms. There may be other differences that I've forgotten already
* Rails 5.2 returns to putting default DOM ids, but still doesn't put the HTML class on forms
* `form_with` does submits via unobtrusive Javascript, so `form` tags are marked with `data-remote="true"`.

https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a is an easy-to-read summary of differences between `form_for` and `form_with`.

Given the variety of expected outputs from different combinations of versions and helpers (`form_for` vs. `form_with`), it was decided to make this PR do the minimum to add `bootstrap_form_with`. Additional testing may be added in future PRs.

Fixes #326.


test/bootstrap_form_with_radio_button_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    #
+    #   assert_equivalent_xml expected, @form_with_builder.collection_radio_buttons(:misc, collection, :id, lambda { |a| a.street.reverse })
+    # end
+    #
+    # test 'collection_radio_buttons renders multiple radios with value defined by lambda correctly' do
+    #   collection = [Address.new(id: 1, street: 'Foo'), Address.new(id: 2, street: 'Bar')]
+    #   expected = %{<div class="form-group"><label class="form-control-label" for="user_misc">Misc</label><div class="radio"><label for="user_misc_address_1"><input id="user_misc_address_1" name="user[misc]" type="radio" value="address_1" /> Foo</label></div><div class="radio"><label for="user_misc_address_2"><input id="user_misc_address_2" name="user[misc]" type="radio" value="address_2" /> Bar</label></div></div>}
+    #
+    #   assert_equivalent_xml expected, @form_with_builder.collection_radio_buttons(:misc, collection, lambda { |a| "address_#{a.id}" }, :street)
+    # end
```
GBH: commented out tests?

README.md
```diff
@@ -40,6 +40,161 @@ Then require the CSS in your `application.css` file:
...
+    <span class="form-text text-muted">A good password should be at least six characters long</span>
+  </div>
+  <div class="form-check">
+    <label class="form-check-label">
+      <input name="user[remember_me]" type="hidden" value="0" />
+      <input class="form-check-input" type="checkbox" value="1" name="user[remember_me]" /> Remember menu
+    </label>
+  </div>
+  <input type="submit" name="commit" value="Log In" class="btn btn-secondary" data-disable-with="Log In" />
+</form>
```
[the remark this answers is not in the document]
lcreid: "Stand-alone"  `bootstrap_form_with` documentation should be easier to update when `bootstrap_form_for` goes away. Also, users often appreciate examples they can just cut and paste.

README.md
```diff
@@ -40,6 +40,161 @@ Then require the CSS in your `application.css` file:
...
+  <input name="utf8" type="hidden" value="&#x2713;" />
+  <div class="form-group">
+    <label class="form-control-label">Email</label>
+    <input value="name@example.com" class="form-control" type="email" name="email" />
+  </div>
+  <input type="submit" name="commit" value="Save " class="btn btn-secondary" data-disable-with="Save " />
+</form>
+```
+
+#### Important Differences Between `form_with` and `form_for`
```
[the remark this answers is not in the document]
lcreid: Some users may appreciate a summary of the differences/tricks about using `form_with` versus `form_for`. If everyone things this is overly redundant or pedantic, by all means I'll remove it.

README.md
```diff
@@ -40,6 +40,161 @@ Then require the CSS in your `application.css` file:
...
+and `bootstrap_form_tag`.
+
+##### Ajax by Default
+`form_with` defaults to submitting forms via Javascript XHR calls,
+like `form_for` or `form_tag` would do if you specified `remote: true`.
+If you want the browser to submit the request
+the same way `form_for` and `form_tag` would do by default,
+you need to specify `local: true` as an option to `form_with`.
+
+##### No Default DOM IDs
```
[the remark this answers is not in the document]
lcreid: Thanks for catching this. I saw that the Rails PR had been merged, but in my rush to submit this PR, I didn't fix the README.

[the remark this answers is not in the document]
lcreid: I've fixed up the README. I've also written some test cases to make sure that `skip_default_ids: false` works as advertised, but I can't test them until I can run Rails 5.2. Will I be able to do that after #366 is merged, or do I need to figure out how to test against Rails 5.2 myself?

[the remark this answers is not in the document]
GBH: You can just temporarily add `gem "rails", "5.2.0.beta3"` in Gemfile to test it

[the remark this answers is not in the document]
lcreid: Thanks, @GBH . I could only find beta2 on rubygems.org, but that's good enough for me for now.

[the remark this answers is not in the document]
lcreid: I need to slow down and read stuff instead of jumping to conclusions. The commit @GBH referenced (https://github.com/rails/rails/commit/d3893ec38ec61282c2598b01a298124356d6b35a) is not the commit I thought it was. I will have to rethink this. The commit makes Rails 5.2+ easier to work with, but I'll have to think about how to document Rails 5.1 and `form_with`.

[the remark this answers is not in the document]
lcreid: It looks like every `form_with` test may have to have a 5.1 and 5.2+ version. There's already too much duplication to test `form_with` and `form_for`. I'll think about this, but if anyone has any suggestions, I'd love to hear them.

tl;dr of the problem: `form_with` in Rails 5.1 doesn't generate any DOM IDs in the HTML. It looks like `form_with` in Rails 5.2 is going back to generating DOM IDs in the HTML by default.

test/bootstrap_form_with_radio_button_test.rb
lcreid: Thanks for catching this. I will uncomment, but I will also have to figure out what to do about the DOM IDs mentioned above.

README.md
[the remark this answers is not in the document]
GBH: The thing is, you don't need to duplicate every single test there is. It's enough to make one for textfield and assume that all others work the same.

lcreid:
I can get this into shape so that it works with Rails <= 5.1 today, but it will fail on Travis on Rails 5.2.0beta2. For Rails 5.2, I will have to re-work the way the tests work. (Thanks to @GBH for stimulating my thinking.) However, Rails 5.2 has some implications for the tests, and for the documentation, that mean I can't promise to have this PR ready for 5.2 before the end of the weekend at the earliest.

Let me know if you want to merge this when it's working for 5.1, or just wait until it also works for 5.2.0beta2. I will arrange my commits so that we can do 5.1 separately if you want.

lcreid:
At the moment, I'll confess I'm struggling a little to fix the `form_with` PR #369. (A paying customer came along with some urgent deadlines. Go figure.) I'm trying to eliminate as much redundancy in the test cases as I can, while still providing adequate coverage. There is more need to test a variety of cases than one might think if they haven't actually tried it. 

I'm hoping I can get some time this weekend. Let me know if I'm holding  things up and we can talk about options.

GBH:
@lcreid is it ok if I take a stab at it? I'm think I can have PR for tomorrow.

lcreid:
@gbh Thanks for the offer, but I think it's mostly an effort to make the testing less repetitive. It's not as simple as it seems. Actually, the actual code changes are not huge, but ensuring adequate test coverage for all the variations of behaviour of the different flavours or forms helpers and versions of Rails that we will be supporting in the future, is an interesting challenge. I would like to finish this, as I've put a lot of effort into it, already.

lcreid:
@mattbrictson what do you think if I remove the testing I'm doing from this PR, so we can merge `form_with`? Then I'll work on new PRs to bring the testing up to where I think it should be. The code as it stands works well up to Rails 5.1. I'm confident that with the testing I would later add, I would find a few issues, but we can deal with them as part of the release cycle for the Bootstrap 4 version of `bootstrap_form`.

GBH:
@lcreid You're over complicating it. All that needs to be tested is the fact that `builder: BootstrapForm::FormBuilder` is getting passed along. You're not re-testing how all form helpers render. And you're most definitely not re-testing how `form_with` works. Here's what tests you'd need to have:

- test `bootstrap_form_with` with default options and single text field
- test `bootstrap_form_with` with all possible **bootstrap** options that can be passed-in
- test `bootstrap_form_with` with default option and single text field for Rails 5.1 as we know the `id` issue there
- repeat same set of tests for `form_with`

Right now I see copy-pasted and edited tests from `form_for`. That's a big 👎  from me.

lcreid:
Thanks for the feedback. I don't think this is the best forum to discuss what is adequate testing, Like I said, I will try to push a new version of this PR without much testing, and then I can work away to my heart's content on more testing while the rest of you push on. OK?

lcreid:
Sorry for the delay. I've pushed something that has limited testing of the new functionality. I still have to add the `fields` helper (`form_with` equivalent of `fields_for`). I would like to provide a bit more of the reasoning, but I'm going out to the Ruby meetup tonight. I look forward to your comments.

lcreid:
Oops. I didn't think to test Rails 5.0. I'll get on that in the morning.

lcreid:
I was getting a failure of Minitest using Rails 5.0, so I pinned Minitest to exactly 5.10.1 (https://github.com/seattlerb/minitest/issues/689). I hope that's okay.

I would like to write up something about how to test. I'm thinking a "Testing Tips" section in the CONTRIBUTING document would be the place to put it. I'm open to other suggestions. I'll do that as a separate PR so we don't muddy any other waters.

test/test_helper.rb
```diff
@@ -88,4 +95,34 @@ def assert_equivalent_xml(expected, actual)
...
+  def remove_default_ids_for_rails_5_1(expected)
+    return expected unless Gem::Version.new('5.1.0') <= Gem::Version.new(::Rails::VERSION::STRING) && Gem::Version.new(::Rails::VERSION::STRING) < Gem::Version.new('5.2.0.beta2')
+    root = Nokogiri::XML(expected)
+
+    # There are more elegant ways to do this, I'm sure, but later for that.
+    root.traverse do |node|
+      node.delete("id") if node.has_attribute?("id")
+    end
+    root.to_s
+  end
```
GBH: Trying to understand what these assertion helpers are for. Why `assert_equivalent_xml` helper is not enough?

README.md
```diff
@@ -40,6 +40,138 @@ Then require the CSS in your `application.css` file:
...
+to the form.
+If you attached styling to the DOM class that `form_for` added to the form element,
+you'll have to add your own code to attach the appropriate class.
+
+#### Nested Forms with `bootstrap_form_with`
+
+This hasn't been tested yet.
+
+### Rails < 5.1
+
```
GBH: Some of this is documentation of `form_with`. Readme should contain info relevant to the library, and not what it wraps. Nobody wants to read a wall of text. Duplicating rendered html produced by bootstrap_form_for and bootstrap_form_with is redundant.

test/documentation_test.rb
```diff
@@ -0,0 +1,140 @@
...
+        f.email_field(:email)
+         .concat(f.password_field(:password, id: :password))
+         .concat(f.submit("Log In"))
+      end
+      #  .concat(f.check_box(:remember_me, id: :remember)) pending #357 place above "Log In"
+
+      assert_equivalent_xml expected, actual
+    end
+  end
+end
```
GBH: Why not just have tests that test something specific instead of shotgun/integration approach here? Don't think this entire test file is needed.

test/test_helper.rb
```diff
@@ -16,6 +16,12 @@ class ActionView::TestCase
   def setup_test_fixture
     @user = User.new(email: 'steve@example.com', password: 'secret', comments: 'my comment')
     @builder = BootstrapForm::FormBuilder.new(:user, @user, self, {})
+    # Simulate how the builder would be called from `form_with`.
+    if '5.1.0' <= ::Rails::VERSION::STRING && ::Rails::VERSION::STRING < '5.2.0'
+      @form_with_builder = BootstrapForm::FormBuilder.new(:user, @user, self, { skip_default_ids: true })
+    else
+      @form_with_builder = BootstrapForm::FormBuilder.new(:user, @user, self, {})
+    end
```
GBH: You can spin up 5.1 form_with builder in the test that actually needs it instead of doing this in the setup.

test/test_helper.rb
lcreid: I have a dream that we can implement one or two helpers that cover the range of variations in output from the Rails helpers across Rails versions. Here's a summary of differences that I'm aware of:

* Prior to Rails 5, tabs were output in alphabetical order. After, they weren't
* Some helpers starting with Rails 5 produced more output, e.g. an empty hidden input on file tags (The genesis of `assert_equivalent_xml` is found in #273 and #278, which specifically addressed these first two issues.)
* Rails 5.1. stopped generating default DOM ids in the form helpers, and stopped putting the `new_class` HTML class on forms. There may be other differences that I've forgotten already
* Rails 5.2 returns to putting default DOM ids, but still doesn't put the HTML class on forms
* Given this history, it's reasonable to expect that future versions of Rails will make other changes that would affect `bootstrap_form`

Some of the guidelines I'm using for testing are:

* Tests should run reasonably quickly. I have a fast laptop but it's already almost two years old, and it runs all the tests in just over a second, so I think we're a ways from having a test suite that runs too long
* Tests should cover all variations in Rails versions that `bootstrap_form` currently supports
* Tests should cover all the variations of input to the gem. For a gem this big, that's of course impossible, but covering more will always be better than covering less
* Do the above in a way that's easy for current and future contributors. This means:
    * Try to concentrate the version handling in as few places as possible
    * Make it just as easy to use the helpers to test multiple versions, as it is to test one version

GBH: Still do not understand the need to mess with `assert_equivalent_xml` by introducing helpers that change passed-in html. Tests themselves should send proper html that should be asserted. To illustrate:

```
# base test
assert expected_html, assert_equivalent_xml(form_for_output)

# good test
assert different_html, assert_equivalent_xml(form_with_output)

# what you're seemingly are doing
assert expected_html, assert_equivalent_xml(mangle_html(form_with_output))
```

At least that's what I think is happening.

mattbrictson:
@lcreid I understand what you are aiming for with the testing approach, but it is a lot to do in a single PR along with all the other form_with stuff. I think I'd rather start by:

1. Introducing `bootstrap_form_with`
2. Adding your documentation to the README
3. Include a bare-minimum test for Rails 5.1 (e.g.`bootstrap_form_with` with a single text field)
4. Likewise, make a similar test for Rails 5.2 (perfect DRY-ness is not a goal in this first pass)

I'm comfortable hacking with git, so I am happy to patch together a new branch that contains the existing work you've done for all of those and retains your authorship of the commits.

I would argue that this first pass might almost be "good enough" to ship for 4.0.0 alpha; we'll probably need a follow-up PR to add `fields_with`. Then we can iterate on testing, `CONTRIBUTING.md` additions, etc. with future PRs.

lcreid:
@mattbrictson that's exactly my thinking. Sorry if my comments are a bit all over the place. In fact, I'd argue that this PR does what you're suggesting, without any more hacking on your part. I'm working on resolving the conflicts from this morning's merges. I had a small problem that I think is due to the Vagrant box I'm using for this work. (I could push something right now, but I'm not 100 percent sure it would pass Travis). I'll ping you shortly with an update on the status.

lcreid:
Sorry. Every time I pull in a merge, I've got to deal with a broken test. In order to do so properly, I need a bit more time. I hope to push something by end of day Pacific Time.

lcreid:
I pushed a conflict-free version, but it should fail on Travis. I expect one failure on all versions of Rails, and an additional failure only on Rails 5.1.4. @mattbrictson I'd appreciate your help, and I will try to look at this later today.

lcreid:
I had a chance to check, and those are the failures I expected. I'll be back on this later tonight.

lcreid:
This is ready for review prior to merge. As discussed above, I will continue to work on enhancing the testing of `form_with` for separate pull requests.

README.md
lcreid: I agree that there's a lot of HTML in the README. I was just following the outline of the documentation for `form_for`.

test/documentation_test.rb
lcreid: The intent of this file to test that what we're putting in the documentation, is in fact what the helpers produce.

lib/bootstrap_form/form_builder.rb
```diff
@@ -314,6 +325,7 @@ def required_attribute?(obj, attribute)
     def form_group_builder(method, options, html_options = nil)
       options.symbolize_keys!
       html_options.symbolize_keys! if html_options
+      options[:id] = html_options[:id] if html_options && @options[:skip_default_ids]
```
mattbrictson: Could you explain why this line is needed?

lib/bootstrap_form/form_builder.rb
```diff
@@ -370,13 +382,19 @@ def form_group_builder(method, options, html_options = nil)
...
       options
     end
 
     def generate_label(id, name, options, custom_label_col, group_layout)
-      options[:for] = id if acts_like_form_tag
+      # id is the caller's options[:id] at the only place this method is called.
+      # The options argument is a small subset of the options that might have
+      # been passed to generate_label's caller, and definitely doesn't include
+      # :id.
+      options[:for] = id if acts_like_form_tag || id
```
mattbrictson: I am also having trouble figuring out this part. If `id` is present, then it will be assigned regardless of whether `acts_like_form_tag` is true or false. Why is the `acts_like_form_tag` check needed at all, then?

Gemfile
```diff
@@ -11,7 +11,7 @@ end
 
 group :test do
   # can relax version requirement for Rails 5.2.beta3+
-  gem "minitest", "~> 5.10.3"
+  gem "minitest", "5.10.3"
```
mattbrictson: I could be wrong, but since 5.10.3 is the only version that matches the `~> 5.10.3` constraint, this change has no actual effect, right?

README.md
```diff
@@ -13,7 +13,7 @@ Bootstrap v4-style forms into your Rails application.
 ## Requirements
 
 * Ruby 2.3+
-* Rails 5.0+
+* Rails 5.0+ (Rails 5.1+ for `bootstrap_form_with`)
```
mattbrictson: 👍

README.md
```diff
@@ -40,6 +40,138 @@ Then require the CSS in your `application.css` file:
 
 ## Usage
 
+### Rails >= 5.1
```
mattbrictson: I do not think we should go as far as recommend `form_with` for Rails 5.1. IMHO the `form_with` released in 5.1 is fundamentally broken as it does not include IDs. I personally would wait until 5.2 to make the leap from `form_for` to `form_with`.

Perhaps we could re-title this section with a caveat:

```
### `bootstrap_form_with` (Rails 5.1+)

Note that `form_with` in Rails 5.1 does not add IDs to form elements and labels by default, which are both important to Bootstrap markup. This behavior is corrected in Rails 5.2.
```

And place it *after* the section where we describe:

```
### `bootstrap_form_for` (Rails 5.0+)
```

README.md
```diff
@@ -40,6 +40,138 @@ Then require the CSS in your `application.css` file:
...
+  </div>
+  <div class="form-check">
+    <label class="form-check-label" for="user_remember_me">
+    <input name="user[remember_me]" type="hidden" value="0" />
+    <input class="form-check-input" type="checkbox" value="1" name="user[remember_me]" /> Remember me</label>
+  </div>
+  <input type="submit" name="commit" value="Log In" class="btn btn-secondary" data-disable-with="Log In" />
+</form>
+```
+
```
mattbrictson: I think it is sufficient to stop here, with a note that says that both `model:` and `url:` styles are supported. We don't need to reproduce all the Rails documentation. (Perhaps a link to the appropriate section of the Rails doc would be nice.) In particular, I would remove these sections as well:

* Important Differences Between `form_with` and `form_for`
* Ajax by Default
* No Default DOM IDs
* Use `fields` Instead Of `fields_for` In Nested Forms
* No Default Classes

README.md
```diff
@@ -84,6 +216,14 @@ If your form is not backed by a model, use the `bootstrap_form_tag`. Usage of th
 <% end %>
 ```
 
+### Future Compatibility
+
+The Rails team has suggested that `form_for` and `form_tag`
+may be deprecated and then removed in future versions of Rails.
+`bootstrap_form` will continue to support
+`bootstrap_form_for` and `bootstrap_form_tag`
+as long as Rails supports `form_for` and `form_tag`.
```
mattbrictson: This is a great clarification 👍 

Could you remove the hard line wrap?

test/documentation_test.rb
mattbrictson: I see what you are going for with this test. As Bootstrap and Rails change, the examples in the README will become out of date, which potentially misleads or confusion people learning about the gem. It also reflects poorly on the project if the README is inaccurate.

However I'm not convinced this is the best solution. I'd like to remove this from this PR and save it for future consideration.

test/bootstrap_radio_button_test.rb
```diff
@@ -14,7 +14,7 @@ class BootstrapRadioButtonTest < ActionView::TestCase
         </label>
       </div>
     HTML
-    assert_equivalent_xml expected, @builder.radio_button(:misc, '1', label: 'This is a radio button')
+    assert_with_builder_radio(expected, :misc, '1', label: 'This is a radio button')
```
mattbrictson: The fact that this test file has to change at all as a result of adding `form_with` indicates to me that there is a little too much going on in this PR. Adding one new feature (`bootstrap_form_with`) should mean adding one new set of tests. It should not require existing tests to be rewritten. I would prefer to see these tests left alone and the `assert_with_builder_radio` and `assert_with_builder` helpers excluded from this PR.

test/bootstrap_fields_test.rb
```diff
@@ -245,11 +257,48 @@ class BootstrapFieldsTest < ActionView::TestCase
...
+      end
+      assert_equivalent_xml remove_default_ids_for_rails_5_1(expected.gsub(/ class="new_user" id="new_user"/, "")),
+                            output
+    end
+  end
+
+  # This test demonstrates that the Rails `fields` method passes its options
+  # to the builder, so there is no need to write a `bootstrap_form` helper
+  # for the `fields` method. If this test fails, it probably means you have to
+  # write a `fields` helper in `lib/bootstrap_form/form_builder.rb`.
```
mattbrictson: It seems this test was written to investigate Rails behavior, and is now no longer needed. Would you agree?

test/bootstrap_fields_test.rb
```diff
@@ -201,6 +202,17 @@ class BootstrapFieldsTest < ActionView::TestCase
       </form>
     HTML
     assert_equivalent_xml expected, output
+
+    ## TODO: DRY up the tests for the _with version
+    if Gem::Version.new(::Rails::VERSION::STRING).release >= Gem::Version.new('5.1.0')
```
mattbrictson: Unless we have reason to believe that `fields` works fundamentally differently than `fields_for`, then I am not sure there is much value for the new tests in this file.

test/bootstrap_form_test.rb
```diff
@@ -12,6 +12,11 @@ class BootstrapFormTest < ActionView::TestCase
       </form>
     HTML
     assert_equivalent_xml expected, bootstrap_form_for(@user) { |f| nil }
+    if Gem::Version.new(::Rails::VERSION::STRING).release >= Gem::Version.new('5.1.0')
+      # TODO: Could argue that we should test without `local: true`
+      assert_equivalent_xml remove_default_ids_for_rails_5_1(expected.gsub(/ class="new_user" id="new_user"/, "")),
+                            bootstrap_form_with(model: @user, local: true) { |f| nil }
```
mattbrictson: I think this is really the only test we need for this PR. But I would split it into 2 tests: one for Rails 5.1, and a second for Rails 5.2+. I prefer it be written as two separate, *new* tests (i.e. not bundled under `test "default-style forms"`). And they should each have their own expected HTML rather than mutating a single expectation with `remove_default_ids_for_rails_5_1`. I believe repeating the HTML makes it much more explicit as to what is being tested and what is expected, which is very important value for tests.

test/bootstrap_fields_test.rb
lcreid: I had the same thought, but then I thought it would also serve to test the assumption going forward and give us a warning if something changes in a future Rails version that the assumption is no longer valid. I think there's value to leave it in (perhaps with a better comment to make the intent clearer), but if you want it removed I'll certainly do so.

Gemfile
lcreid: Sorry. This was part of some flailing I did around the Minitest versions, and I shouldn't have let it slip through to the PR. :-(

README.md
lcreid: Yes. I wrote the README relatively early in my work on this PR, and once I actually started to use `form_with` in my own app I was feeling that `form_with` in 5.1 wasn't really ready for prime time.

README.md
lcreid: Back in September, when I was first trying my fork with `bootstrap_form_with`, this section brought together the things I was learning about `form_with` versus `form_for`. But now I'm sure I can find a good link to differences, so I'll take this section out and replace it with a link or two to other documentation.

test/bootstrap_radio_button_test.rb
lcreid: I'll remove all testing for `bootstrap_form_with` and `fields` other than the two cases you mentioned above. I'll continue to work on my testing branch and see if I still feel there's a need for more testing.

I've come around to the idea of making this PR minimal, so that's what I'll do. FWIW, the reasons why I wrote more extensive tests are:

* If `form_with` is someday going to replace `form_for` and `form_tag`, then I thought it would be better write a complete set of `bootstrap_form_with` tests while I had my head in the problem, rather than leaving it to someone else, five years from now, to figure out how to migrate the `bootstrap_form_for` tests to `bootstrap_form_with`
* Back in August I understood very little of this gem, and wanted to make sure I wasn't breaking anything
* When I tried to use `bootstrap_form_with` in my app, lots of stuff broke. Most of it turned out to be due to the differences between `form_for` and `form_with`, but I didn't figure that out until after I wrote some tests for `bootstrap_form_with`
* General distrust of myself any time I think that, "I only need to test this one little difference". :-)

And I really appreciate the feedback, so keep the comments coming!

test/bootstrap_form_test.rb
lcreid: Yes, I'll do that. (And see my next comment.)

lib/bootstrap_form/form_builder.rb
lcreid: This was to address part of the consequence of no default IDs in Rails 5.1 `form_with`. I'll take this out, and let it be handled by #357. And yes, the `acts_like_form_tag` becomes redundant when I added `|| id`. "Doh!" I don't know what I was thinking. :-)

README.md
lcreid: Sorry. An old `nroff` habit.

lib/bootstrap_form/form_builder.rb
lcreid: I'm pretty sure this is part my attempt to handle what is more elegantly handled in #357. The (minimized) tests pass without this line, so I'll take it out.

mattbrictson:
Thanks! I know this PR has been a lot of work so it is great to finally hit the merge button. I am going to squash it down to one commit and then merge.

lcreid:
Thanks. It was a great learning experience for me. I really appreciate your help and patience with me while I ground away on this.

Issue #371, opened 2018-01-11
MiniTest Test Runner Exits After One Test Failure


lcreid:
When running tests, MiniTest exists after one test failure. For example:
```
vagrant@rails-5-1:/vagrant$ bundle exec rake
Run options: --seed 41303

# Running:

............................................F

Failure:
BootstrapCheckboxTest#test_check_box_is_wrapped_correctly [/vagrant/test/bootstrap_checkbox_test.rb:12]:
-<div aclass="form-check">
+<div class="form-check">
   <label class="form-check-label" for="user_terms"><input name="user[terms]" type="hidden" value="0"/><input class="form-check-input" id="user_terms" name="user[terms]" type="checkbox" value="1"/> I agree to the terms</label>
 </div>
\ No newline at end of file



/var/lib/gems/2.3.0/gems/railties-5.1.4/lib/rails/test_unit/reporter.rb:70:in `method': undefined method `test_check_box_is_wrapped_correctly' for class `Minitest::Result' (NameError)
	from /var/lib/gems/2.3.0/gems/railties-5.1.4/lib/rails/test_unit/reporter.rb:70:in `format_rerun_snippet'
	from /var/lib/gems/2.3.0/gems/railties-5.1.4/lib/rails/test_unit/reporter.rb:23:in `record'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:786:in `block in record'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:785:in `each'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:785:in `record'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:334:in `run_one_method'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:321:in `block (2 levels) in run'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:320:in `each'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:320:in `block in run'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:360:in `on_signal'
	from /var/lib/gems/2.3.0/gems/minitest-5.11.1/lib/minitest.rb:347:in `with_info_handler'
[...]

PR #372, opened 2018-01-11
Pin Minitest to work around crash on test failure.

lcreid:
Minitest 5.11 is broken for environments, like Rails, that use custom reporters. Pin to 5.10.3 in the meantime.

Addresses issue #371.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,3 +16,5 @@ group :test do
   gem "sqlite3"
   gem "timecop", "~> 0.7.1"
 end
+
+gem "minitest", "~> 5.10.3"


donv:
Thanks!

mattbrictson:
I could be wrong, but my understanding based on this thread https://github.com/bootstrap-ruby/rails-bootstrap-forms/pull/366#discussion_r160315317 is that whenever we add or pin a dependency we have to repeat it in each of the Gemfiles. I.e. `Gemfile` plus all the ones in `test/gemfiles`.

Also, shouldn't `minitest` go in the `:test` group?

GBH:
Yes. This PR wasn't ready to be merged.

PR #374, opened 2018-01-13
Putting minitest dependency in the right place. Fixing PR#372

GBH:
test gemspecs for rails < 5.2 will have it pinned permanently


mattbrictson:
Thanks. When the next 5.2 beta is released we should be able to remove some of the pins, as you point out.

PR #375, opened 2018-01-13
[v4] Fleshing out test/dummy form

GBH:
Within test/dummy: `rails s` and you'll be able to see how forms are actually getting rendered.

So far I see problem with single checkbox, radio_button. Submit buttons are not aligned correctly. Div containers where they are not supposed to be (old v3 markup). I'm sure there's more. 

This how we can validate that helpers are actually producing markup that works with Bootstrap as you can't really see it by staring at HTML inside test assertions. 

I previously added note about it in the CONTRIBUTING, but maybe it should be more prominent.


mattbrictson:
I'm also happy to merge this as-is for now.

lcreid:
+1 to the idea that we should actually take a look at how this is all rendered before we release the new gem. I've notice some of the problems @gbh mentioned in my app that's using the Bootstrap 4 branch.

GBH:
I like the idea of demo app. Will do needed tweaks.

GBH:
Ok, moved test/dummy to demo.

- Amended demo/README on how to use
- Amended CONTRIBUTING
- Added few more form field types
- Didn't make a rake task as it's basically so you can type 1 line instead of 2. Also I don't want to deal with different OS's on how to open urls in browsers. Somebody can make a separate PR if they feel the need.
- Noticed odd warning you get when running tests after creating development.sqlite3 db. `ActiveRecord::EnvironmentMismatchError` Seems that Rails touches test db for some reason during development db creation. Doesn't affect anything though.

This PR is ready to go.

GBH:
Can this be merged? Want to start on actually fixing things and want this to be in master. Thanks.

mattbrictson:
Thanks for your patience! Looks great 💯

PR #377, opened 2018-01-14
[v4] Removing nested_form support

GBH:
Untested support for a gem that was abandoned 4 years ago. There's a better way to handle dynamic forms these days (like cocoon, perhaps). If somebody still really wants it, they can create wrapper for this in their own app.


mattbrictson:
Good catch! I can't imagine people would be using this abandoned gem, but who knows, maybe there are some forks out there. In any case, I think a mention in the CHANGELOG is warranted here.

GBH:
added entry to changeliog

mattbrictson:
I went ahead and resolved the CHANGELOG conflict. I'll merge once Travis passes.

mattbrictson:
Well, the latest release of `nested_form` was [April 5, 2013](https://rubygems.org/gems/nested_form), and no PRs have been merged since Dec 1, 2013, so I think it fair to say the gem is abandoned. I'd be surprised if it works with the latest version of Rails.

I have used `cocoon` with bootstrap_form in the past; I seem to remember it working well. If you run into any problems using cocoon with bootstrap_form please open an issue here.

PR #378, opened 2018-01-14
[v4] Removing deprecated methods

GBH:
`check_boxes_collection` and `radio_buttons_collection` are gone for a long time now.


mattbrictson:
Great opportunity to remove deprecated methods. Nice find 👍 

Could you also add CHANGELOG entry?

GBH:
Added changelog entry

PR #379, opened 2018-01-15
[v4] Fix file input class

atipugin:
Latest (beta3) version of bootstrap 4 requires file input to have `form-control-file` class. This PR adds this class.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -61,7 +61,8 @@ module BootstrapForm
     end
 
     def file_field_with_bootstrap(name, options = {})
-      form_group_builder(name, options.reverse_merge(control_class: nil)) do
+      options = options.reverse_merge(control_class: 'form-control-file')
+      form_group_builder(name, options) do
         file_field_without_bootstrap(name, options)
       end
     end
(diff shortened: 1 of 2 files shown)


lib/bootstrap_form/form_builder.rb
```diff
@@ -61,7 +61,8 @@ def initialize(object_name, object, template, options)
     end
 
     def file_field_with_bootstrap(name, options = {})
-      form_group_builder(name, options.reverse_merge(control_class: nil)) do
+      options[:control_class] ||= 'form-control-file'
```
mattbrictson: I would prefer `reverse_merge` here, because it creates a new hash instead of modifying the `options` in place.

atipugin:
I'm not sure about `reverse_merge`. It doesn't work if we pass `options.reverse_merge(...)` as argument to `form_group_builder`, because `form_group_builder` modifies new version of `options`, but yielded code (`file_field_without_bootstrap(name, options)`) still references old (unmodified) hash.

atipugin:
Done

atipugin:
Looks like master is broken for now :/

lcreid:
This is the same failing test that I reported in #397.

mattbrictson:
I'll update this PR with the latest from master

mattbrictson:
All green! Thanks for the PR, @atipugin 🙌

PR #380, opened 2018-01-16
[v4] Removing stylesheet

GBH:
Reasons why it's not needed:

- Originally it was there to style required fields. Not anymore.
- Nobody uses default Rails form fields for date/time. There are million timepickers out there.
- If somebody actually uses default datetime fields and form-errors summary they can style it themselves.
- Silly to ship app/assets for 3 lines of css that nobody would need.


mattbrictson:
I'm undecided on this one. Other reviewers?

desheikh:
I actually am using the default datepicker for rails in some of my projects, however I had no idea bootstrap_form set any styling for it.
Somewhat mixed on this, as it's one of those cases where rails forms and bootstrap don't overlap.
I guess we need to decide if this lies within the domain of this project.

lcreid:
Actually, @mattbrictson, I count the majority opinion as being "leave it in if it doesn't hurt anything." I think we're saying we didn't want to block it, but we couldn't see the harm in leaving it in. Sorry if my approval with comment wasn't clear.

mattbrictson:
OK, if there is a mixed opinion on this then the safest course of action is to leave it as-is. Thanks for your comment @lcreid .

@GBH I'm closing this for now, but as always, thanks for the PR.

PR #381, opened 2018-01-16
[v4] Removing old Rails checks

GBH:
Since we don't care about Rails 3 and 4 we don't need those checks (not like any of that stuff was tested anyways)


mattbrictson:
Thanks, nice cleanup 🙌

PR #389, opened 2018-01-18
[v4] Fixing labels in horizontal forms

### LINKED ISSUE #384: [v4] Labels for horizontal forms are lacking `col-form-label` class.
Right now I see:

```html
<label class="col-sm-2 required" for="user_email">
```

when it needs to be 

```html
<label class="col-sm-2 col-form-label required" for="user_email">
```

Only horizontal layout needs this. I think somebody refactored it out instead of changing it to a proper class.

For reference: https://getbootstrap.com/docs/4.0/components/forms/#horizontal-form

GBH:
Fixes #384


mattbrictson:
Good catch! Thanks for another solid PR 👍

PR #390, opened 2018-01-18
[v4] Fixing test invalidation statement

GBH:
Just a quick tweak for clarity


mattbrictson:
Makes sense. Thanks! 💯

Issue #391, opened 2018-01-18
[v4] assert_equivalent_xml helper doesn't work as expected


GBH:
```ruby
test "assert_equivalent_xml helper" do
  expected = %{<header>paragraph</header><main>main</main>}
  assert_equivalent_xml expected, %{<header>paragraph</header><main>main</main>}

  # This should fail, but doesn't
  expected = %{<header>paragraph</header>}
  assert_equivalent_xml expected, %{<header>paragraph</header><main>main</main>}

  # This should fail, but doesn't
  expected = %{<header>paragraph</header><main>main</main>}
  assert_equivalent_xml expected, %{<header>paragraph</header>}
end

```
I'd assume that should fail? Seems that it's ignoring all elements after the first root one.

GBH:
K. Issue is that assertion helper is using `Nokogiri::XML` to load HTML. `<header>foo</header><footer>bar</footer>` is not a valid XML. So it just ignores all nodes after the first one. Using `Nokogiri::HTML` crams `<html><body>` everywhere and messing things up. I think the easiest solution is to wrap everything in `<test> .. </test>` tag within the helper to ensure proper XML.

Will make PR.

lcreid:
Thanks for this explanation. It makes everything much clearer. Probably explains some strange behaviour I've been observing all along. And thanks for this fix!

PR #392, opened 2018-01-18
[v4] Fixing assert_equivalent_xml helper

GBH:
As expected a bunch of tests weren't actually testing anything. Related to #391


GBH:
@mattbrictson can you please prioritize this PR? It's a bit of a blocker.

mattbrictson:
Great catch and an elegant fix. Thanks!

Merged via effe4f2

PR #338, opened 2017-08-01
Fixed offset syntax for Bootstrap 4

SergeyBurtsev:
Changed column's offset classes to reflect Bootstrap 4 syntax.
https://v4-alpha.getbootstrap.com/layout/grid/#offsetting-columns

PR #393, opened 2018-01-19
[v4] Submit button

GBH:
- defaulting to `btn` class. Anything else is changed via `class:` option. 
- allowing block content for additional html like cancel links
- automatically is put in column if used in a `layout: horizontal` form
- fixed offset class generation (closes #338)

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -207,6 +207,7 @@ module BootstrapForm
 
     bootstrap_method_alias :collection_radio_buttons
 
+    # TODO: Needs documention
     def form_group(*args, &block)
       options = args.extract_options!
       name = args.first
@@ -245,11 +246,10 @@ module BootstrapForm
 
     bootstrap_method_alias :fields_for
 
-    # Add bootstrap formatted submit button. It's a primary button by default.
-    # If you need to change its type or add another css class, you need to
-    # override all css classes like so:
+    # Add bootstrap formatted submit button. If you need to change its type or
+    # add another css class, you need to override all css classes like so:
     #
-    #   <%= form.submit class: "btn btn-secondary custom-class" %>
+    #   <%= form.submit class: "btn btn-info custom-class" %>
     #
     # You may add additional content that directly follows the button. Here's
     # an example of a cancel link:
@@ -259,10 +259,17 @@ module BootstrapForm
     #   <% end %>
     #
     def submit(value = nil, options = {}, &block)
-      options.reverse_merge!(class: "btn btn-primary")
-      out = super(value, options)
+      out = super(value, options.reverse_merge(class: "btn"))
       out += capture(&block) if block_given?
-      out
+
+      form_group do
+        out
+      end
+    end
+
+    # Same as submit button, only with btn-primary class added
+    def primary(value = nil, options = {}, &block)
+      submit(value, options.reverse_merge(class: "btn btn-primary"), &block)
     end
 
     private
@@ -280,7 +287,7 @@ module BootstrapForm
     end
 
     def offset_col(label_col)
-      label_col.sub(/^col-(\w+)-(\d)$/, 'col-\1-offset-\2')
+      label_col.sub(/^col-(\w+)-(\d)$/, 'offset-\1-\2')
     end
 
     def default_control_col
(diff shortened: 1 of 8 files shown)


lib/bootstrap_form/form_builder.rb
```diff
@@ -245,6 +245,26 @@ def fields_for_with_bootstrap(record_name, record_object = nil, fields_options =
...
+    # override all css classes like so:
+    #
+    #   <%= form.submit class: "btn btn-secondary custom-class" %>
+    #
+    # You may add additional content that directly follows the button. Here's
+    # an example of a cancel link:
+    #
+    #   <%= form.submit do %>
+    #     <%= link_to "Cancel", "/", class: "btn btn-link" %>
+    #   <% end %>
```
mattbrictson: Nice docs. 💯

lib/bootstrap_form/form_builder.rb
```diff
@@ -245,6 +245,26 @@ def fields_for_with_bootstrap(record_name, record_object = nil, fields_options =
...
+    #
+    def submit(value = nil, options = {}, &block)
+      options.reverse_merge!(class: "btn btn-primary")
```
mattbrictson: I prefer not to mutate the method parameters. How about this?

```ruby
options = options.reverse_merge(class: "btn btn-primary")
```

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -1,15 +1,6 @@
 module BootstrapForm
   module Helpers
     module Bootstrap
-      def submit(name = nil, options = {})
-        options.reverse_merge! class: 'btn btn-secondary'
-        super(name, options)
-      end
-
-      def primary(name = nil, options = {})
```
mattbrictson: Is this removal a breaking change?

GBH: I didn't remove it

lcreid:
Is there a chance the extra markup will change the layout of pages being migrated from v2.7? Could you try (in the demo app or Bootply) the Bootstrap 3 and Bootstrap 4 `submit` in all three form layouts (`:default`, `:inline`, `:horizontal`)?

GBH:
You mean `form-group` wrapper? It doesn't affect how inline or vertical forms render.

lib/bootstrap_form/form_builder.rb
```diff
@@ -245,6 +246,32 @@ def fields_for_with_bootstrap(record_name, record_object = nil, fields_options =
...
+    #
+    def submit(value = nil, options = {}, &block)
+      out = super(value, options.reverse_merge(class: "btn"))
+      out += capture(&block) if block_given?
+
+      form_group do
+        out
+      end
+    end
+
```
lcreid: Would it make sense to define the `submit_without_bootstrap` method so the user can choose to do their own button, considering the additional functionality that this PR introduces?

lib/bootstrap_form/form_builder.rb
```diff
@@ -245,6 +246,32 @@ def fields_for_with_bootstrap(record_name, record_object = nil, fields_options =
...
+    #
+    #   <%= form.submit do %>
+    #     <%= link_to "Cancel", "/", class: "btn btn-link" %>
+    #   <% end %>
+    #
+    def submit(value = nil, options = {}, &block)
+      out = super(value, options.reverse_merge(class: "btn"))
+      out += capture(&block) if block_given?
+
+      form_group do
```
lcreid: Sorry my earlier comment wasn't complete. If we're adding a `div.form-group` to the markup for a submit button, this may or may not change the layout of some pages. I ran into a lot of cases where the addition of even one `div` would change the way the browser actually renders Bootstrap 4 markup. I resorted to trying just about everything in Bootply, comparing the Bootstrap 3/`bootstrap_form` v2.7 rendering to the Bootstrap 4/`bootstrap_form` v4.0 rendering. Tedious, but might avoid some surprises.

lib/bootstrap_form/form_builder.rb
GBH: You can just do `submit_tag` to bypass builder. Same effect.

lib/bootstrap_form/form_builder.rb
GBH: There's a /demo here where you can try out all types of scenarios. Easier than copy-pasting code into Bootply.

PR #401, opened 2018-01-23
Miscellaneous cleanup; rewrite CHANGELOG for 4.0.0

mattbrictson:
This PR includes a handful of tweaks:

1. Fix a mistake in Ruby version requirement listed in the README: we support 2.2.2+, not 2.3+.
2. Using the various gemfiles in `test/gemfiles` causes `.lock` files to be created. These should never be committed, so add them to the `.gitignore`.
3. The repo URL has recently changed; update the homepage in the gemspec to match.
4. The minitest fix in Rails has been backported to the 5.1 branch; add a comment as a reminder to remove our workaround once 5.1.5 has been released.
5. The CHANGELOG did not explain the magnitude of the changes since the last bootstrap_form release (2.7.0) and there are too many markup changes to reasonably list. Rewrite the CHANGELOG to explain what has changed at a high-level and direct people to the Bootstrap v4 docs for a more complete explanation of the markup changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ demo/tmp/
 .rbenv-gemsets
 *.swp
 Gemfile.lock
+test/gemfiles/*.lock
 .ruby-version
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ demo/tmp/
 Gemfile.lock
 test/gemfiles/*.lock
 .ruby-version
+Vagrantfile
+.vagrant
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,29 @@
 ## [Pending Release][]
 
-Bugfixes:
-  - Your contribution here!
-  - form-control-danger is replaced with is-invalid for bootstrap 4.0.0.beta3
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -8,4 +8,5 @@ demo/tmp/
 .rbenv-gemsets
 *.swp
 Gemfile.lock
+test/gemfiles/*.lock
 .ruby-version
```
lcreid: There's a `test/dummy` directory on my machine right now that seems to contain only databases and log files. Could we add this to `.gitignore`. Also, for those of us who use Vagrant, would you mind adding `.vagrant` and `Vagrantfile` to `.gitignore`?

mattbrictson: Sure, I'll add the Vagrant stuff. As far as `test/dummy` goes, that directory was a byproduct of the old location of the test Rails app, which as since moved to `demo/`. You can safely remove the `test/dummy` directory and those files will not come back.

mattbrictson:
Thanks for the review, @lcreid. I'll make the gitignore changes and figure out how to make @bootstrap-ruby-bot happy.

PR #403, opened 2018-01-24
Fix EnvironmentMismatchError when running tests

### LINKED ISSUE #402: After running demo app, tests print EnvironmentMismatchError warning
If I run the demo app using these steps as listed in CONTRIBUTING.md:

```
cd demo
rake db:schema:load
rails server
```

And then later I run the tests, I then see

```
$ rake
rails aborted!
ActiveRecord::EnvironmentMismatchError: You are attempting to modify a database that was last run in `development` environment.
You are running in `test` environment. If you are sure you want to continue, first set the environment using:

        bin/rails db:environment:set RAILS_ENV=test

bin/rails:4:in `require'
bin/rails:4:in `<main>'
Tasks: TOP => db:test:load => db:test:purge => db:check_protected_environments
(See full trace by running task with --trace)
```

The tests otherwise continue to run normally, without errors.

I have gotten the error go away by doing this:

```
rm -rf demo/db/*.sqlite3
```

But that doesn't seem ideal. The tests should just work without scary warnings.

mattbrictson:
Once demo app is used, then subsequent runs of `rake test` would print an `EnvironmentMismatchError`. Removing the `RAILS_ENV=test` assignment fixes this warning.

Fixes #402

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -4,8 +4,6 @@ require 'nokogiri'
 require 'equivalent-xml'
 require 'mocha/mini_test'
 
-ENV["RAILS_ENV"] = "test"
-
 require_relative "../demo/config/environment.rb"
 require "rails/test_help"
 


mattbrictson:
Thanks for the review!

lcreid:
@mattbrictson I've confirmed that if you roll back this PR and do:
```
rm demo/db/test.sqlite3
```
You can run your tests without errors or the annoying environment mismatch error. It looks like the annoying message is actually caused by a bug in Rails 5.0 and 5.1, the fix for which has been backported to 5.0 and 5.1 but not yet released.

How about if I document two work-arounds in issue #402, and we reverse this PR? I'm open to other suggestions.

P.S. is there an easy way for me to actually test a PR like this before I run it? Easier than forking your repo, (which isn't that hard, really)? (In this case, I should have just done the change manually, but I'm asking for the future.)

lcreid:
Thanks for the tips. I didn't know about `hub`. I'll give it a try.

Issue #399, opened 2018-01-23
Pretty-print actual output when HTML comparison fails, for more precise diff


mattbrictson:
When a test using `assert_equivalent_xml` fails, the expected HTML is nicely indented, but the actual HTML is not. This results in a diff that is hard to interpret, as shown here:

```diff
-<div class="form-group row">
-  <div class="col-sm-10 col-sm-offset-2">Hallo</div>
-</div>
-<div class="form-group row">
-  <label class="col-sm-2 required" for="user_email">Email</label>
-  <div class="col-sm-10">
-    <input class="form-control" id="user_email" name="user[email]" type="text" value="steve@example.com"/>
-  </div>
-</div>
-
+<div class="form-group row"><div class="col-sm-10 col-sm-offset-2">Hallo</div></div><div class="form-group row"><label class="col-form-label col-sm-2 required" for="user_email">Email</label><div class="col-sm-10"><input class="form-control" id="user_email" name="user[email]" type="text" value="steve@example.com"/></div></div>
```

It would be great if _both_ the actual and expected output were nicely indented, so that a more precise diff could be printed, like this:

```diff
   <div class="col-sm-10 col-sm-offset-2">Hallo</div>
 </div>
 <div class="form-group row">
-  <label class="col-sm-2 required" for="user_email">Email</label>
+  <label class="col-form-label col-sm-2 required" for="user_email">Email</label>
   <div class="col-sm-10">
     <input class="form-control" id="user_email" name="user[email]" type="text" value="steve@example.com"/>
   </div>
```

(This may or not be possible using the tools available to us via the `equivalent-xml` gem; I'm not sure.)

mattbrictson:
This seems to do the trick. I'll do more thorough testing and open a PR later this week, unless someone else wants to jump on it first.

```diff
diff --git a/test/test_helper.rb b/test/test_helper.rb
index daf29da..b677f2d 100644
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -53,8 +53,8 @@ class ActionView::TestCase
 
   # Expected and actual are wrapped in a root tag to ensure proper XML structure
   def assert_equivalent_xml(expected, actual)
-    expected_xml        = Nokogiri::XML("<test-xml>\n#{expected}\n</test-xml>")
-    actual_xml          = Nokogiri::XML("<test-xml>\n#{actual}\n</test-xml>")
[...]

PR #404, opened 2018-01-24
Pretty-print output of failed tests #399.

lcreid:
Thanks to @mattbrictson for figuring this out. I just implemented what he suggested. :-)

This PR doesn't affect end users. It simply pretty-prints output from failed tests, so that it's easier to see where the differences are.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -53,8 +53,8 @@ class ActionView::TestCase
 
   # Expected and actual are wrapped in a root tag to ensure proper XML structure
   def assert_equivalent_xml(expected, actual)
-    expected_xml        = Nokogiri::XML("<test-xml>\n#{expected}\n</test-xml>")
-    actual_xml          = Nokogiri::XML("<test-xml>\n#{actual}\n</test-xml>")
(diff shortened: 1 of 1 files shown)


mattbrictson:
Just to be sure: do we have any tests that are sensitive to white space? My understanding is that the `noblanks` option causes blank text nodes to be ignored by the XML parser. So for example, these two snippets would be considered equivalent:

```html
<label>Foo</label><input>
```

```html
<label>Foo</label> <input>
```

I think I'm OK with that, but just wanted to make sure we're in agreement.

mattbrictson:
Also: could you add "Fixes #399" to the description of this PR (i.e. the first comment of this thread)? In general it is a good practice to reference the issue that a PR is fixing using that exact phrasing. That way GitHub will automatically close the referenced issue when the PR is merged.

https://help.github.com/articles/closing-issues-using-keywords/

lcreid:
Do we have tests that are sensitive to white space? I've asked myself the same question. Certainly we don't have any that fail because we're handling white space differently now, as Travis shows. But we might be masking future changes that should be identified as a failure. I can look into that, but probably not in the next few days, so if someone else wants to look at it, go for it.

mattbrictson:
I skimmed through the tests and I am pretty satisfied that the white space won't be a problem. I'll merge this, thanks! 🙌

Issue #395, opened 2018-01-19
[v4] Rendering of checkboxes and radios changed


GBH:
Right now I see:

```html
<div class="form-check">
  <label class="form-check-label" for="user_misc_1">
    <input class="form-check-input" type="checkbox" value="1" name="user[misc][]" id="user_misc_1" />
    Foo
  </label>
</div>

<div class="radio">
  <label for="user_misc_1">
    <input type="radio" value="1" name="user[misc]" id="user_misc_1" />
    Foo
  </label>
</div>
```
Where it should be:

```html
<div class="form-check">
  <input class="form-check-input" type="checkbox" value="" id="defaultCheck1">
  <label class="form-check-label" for="defaultCheck1">
    Default checkbox
  </label>
</div>

<div class="form-check">
  <input class="form-check-input" type="radio" name="exampleRadios" id="exampleRadios1" value="option1" checked>
  <label class="form-check-label" for="exampleRadios1">
    Default radio
  </label>
</div>
```

See: https://getbootstrap.com/docs/4.0/components/forms/#default-stacked

mattbrictson:
Good find. Definitely a bug we need to fix for 4.0.0.

lcreid:
@jeffblake mentioned this in a comment on #361. It's good to track this in its own issue.

lcreid:
Should custom radio buttons be rendered the same as before:
```
      <div class="custom-control custom-radio">
        <input class="custom-control-input" id="user_misc_1" name="user[misc]" type="radio" value="1" />
        <label class="custom-control-label" for="user_misc_1">This is a radio button</label>
      </div>
```
Should custom radio buttons be rendered the same as before:
```
      <div class="custom-control custom-checkbox">
        <input name="user[terms]" type="hidden" value="0" />
        <input class="custom-control-input" id="user_terms" name="user[terms]" type="checkbox" value="1" />
        <label class="custom-control-label" for="user_terms">I agree to the terms</label>
      </div>
```

lcreid:
This is what needs to be done for this issue (moved here from PR #407):

- [x] correcting the classes on the radio button elements (https://github.com/bootstrap-ruby/bootstrap_form/pull/407#issuecomment-361032165)
- [x] inline radios and checkboxes
- [x] custom controls -- Looks like this was already done. Let me know if you see any discrepancies between our output and the Bootstrap 4 documentation
- [ ] update the README where the examples of generated markup are out-of-date
[...]

PR #409, opened 2018-01-28
Use Bootstrap 4 radio button classes.

lcreid:
Use the correct Bootstrap 4 classes for radio buttons.

This PR corrects the classes on inline radio buttons, but does not address the overall format (wrapping) of inline radio buttons for Bootstrap 4. That will be in a separate PR.

There is some redundancy in this code, because I chose to try to minimize changes to achieve the desired functionality, rather than refactor to the DRYest state. I expect that the code will naturally get DRYer in the next PR, and that any remaining DRYing up can be handled by a separate PR later.

This is part of the work for #395.


mattbrictson:
Nice, bite-sized PR. Thanks! 👍

PR #410, opened 2018-01-28
Handle inline radio buttons and check boxes for Bootstrap 4.

lcreid:
The Bootstrap 4 tags for radio buttons and check boxes is more regular, so the helpers are somewhat shorter, at the cost of perhaps making the review more difficult.

If there's a way to show the diff without comparing white space, it might give you a better idea of the real changes, most notably in the test cases.

This PR partially addresses #395.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -170,15 +170,12 @@ module BootstrapForm
           radio_html.concat(label(name, options[:label], value: value, class: ["custom-control-label", label_class].compact.join(" ")))
         end
       else
-        label_class = " #{label_class}" if label_class
-        if options[:inline]
+        wrapper_class = "form-check"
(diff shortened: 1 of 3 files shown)


mattbrictson:
FYI you can append `?w=1` to a PR diff to ignore white space, like this: <https://github.com/bootstrap-ruby/bootstrap_form/pull/410/files?w=1>

lcreid:
Nice catch. I was so into just mechanically changing the existing test cases that I didn't notice that we weren't testing the case where the form has the `layout: inline`. In the spirit of small, bite-sized PRs, I would humbly suggest we merge this one, and raise a new issue for the form-level inline. I won't be putting in an issue tonight, but I can do it in the morning if you haven't already done so.

mattbrictson:
Sounds good! If you could open the issue tomorrow I'd appreciate it.

Issue #412, opened 2018-01-29
Confirm Functioning of Options That Can Be Specified at Form or Form Element Level


lcreid:
Some questions have been raised about options, like `layout: inline`/`inline:true`, that can be specified at the form level (on `bootstrap_form_for`, `bootstrap_form_tag`, or `bootstrap_form_with`), or at the individual element level (on `checkbox` for example). The documentation may be unclear or wrong, the code may be wrong, or both.

This issue is intended to capture the discussion around these options. It may lead to more specific issues and pull requests. Please contribute your thoughts.

From: https://github.com/bootstrap-ruby/bootstrap_form/pull/410#pullrequestreview-92080618:

> Some parts of the README imply that you just add `layout: :inline` to `bootstrap_form_for` and you are done. But that is either broken or the doc is wrong, because as I see in the tests, you must explicitly pass `inline: true` to each radio or checkbox (in addition to passing `layout: :inline` for the form?).

The options that should be looked at are:

- `layout: inline`/`inline:true`
- `layout: horizontal`
- `hide_label`
- `skip_label`
- `label_class`
- `label_col`
- `control_col`
- `inline_errors`
- `label_errors`

Please add any other options that you think should be considered.

lcreid:
#291 requested passing `label_class` to each of the `label`s when `hide_label: true` is given to `collection_check_boxes`. It begs the question as to how to format the label regardless of the state of `hide_label`.

lcreid:
Could someone please confirm that [this](https://www.bootply.com/qPwL3lE3A5) is how we would want check boxes to look given `layout: :inline`, `layout: horizontal`, and default? Make sure to scroll down to see `hide_label` and `skip_label`.

The examples can be taken as what should be generated by `collection_check_boxes` in a simple case. The user could also generate the same effect using `check_box` directly, but would have to provide some additional markup, for the "group" labels (Bootstrap calls them legends), and for the rows in the case of `layout: :horizontal`

In the `layout: :horizontal` case, `label_col` and `control_col` would be used for the label (legend) and control (the check boxes combined with their labels). The ``skip_label` case would not use the default, but would use the whole width for the controls.

(I'll do radio buttons next.)

lcreid:
Please confirm that [this](https://www.bootply.com/Q9HoFkjVNk) is how we would want radio buttons to look given `layout: :inline`, `layout: :horizontal`, and default? Make sure to scroll down to see `hide_label` and `skip_label`.

[...]

Issue #162, opened 2014-11-03
skip_label doesn't work


hdanh:
Hi,

I still get label tag rendered when using skip_label: true, any idea ?

Issue #161, opened 2014-10-22
Ability to create radio button without label?


akolomiychuk:
I want to achieve this html using form_builder
```html
<div class="form-group">
  <label class="sr-only control-label" for="quiz_multiple_question_answer0">Answer0</label>
  <div class="input-group">
    <span class="input-group-addon">A</span>
    <input class="form-control form-control" id="quiz_multiple_question_answer0" name="quiz_multiple_question[answer0]" placeholder="Answer A" type="text">
    <span class="input-group-addon"><input id="quiz_multiple_question_solution_0" name="quiz_multiple_question[solution]" type="radio" value="0"> </span>
  </div>
</div>
```

with following code
```ruby
f.text_field("answer#{i}".to_sym, class: 'form-control', placeholder: "Answer #{letters[i]}", hide_label: true, append: f.radio_button(:solution, i), prepend: "#{letters[i]}")
```

But seems that radio button cannot be created without label. I tested it on version `2.1.1`, but seems that this is actual for current `master` too.

I already monkey-patched it locally(added `no_label: true` option), so I can create a PR if you approve that this is a feature which you want to have.

onebree:
I would like your monkey path code for the `no_label` option. I am having the same option with checkboxes. I have a series of checkboxes inside a table, each column with a heading -- no need for a label. In the mean time, I have `label: ""`, but now I cannot check the box in my specs... (Without selecting the box by a huge id.)

Soullivaneuh:
Got the same issue:

```haml
= f.form_group :active, label: { text: 'Active' } do
  = f.check_box :active, hide_label: true
```

Give me this:

![checkbox_nolabel_not_expected](https://cloud.githubusercontent.com/assets/1698357/6963765/1155d110-d943-11e4-9a0a-592b4a41a8a8.png)

But I expected this:

![checkbox_nolabel_expected](https://cloud.githubusercontent.com/assets/1698357/6963769/1575bab2-d943-11e4-924b-398a8fbaeb02.png)

@akolomiychuk I would be happy to see this PR. But why not using the `hide_label` option instead of `no_label`?

onebree:
I agree, I would rather have `hide_label`, because it is still present in the document, for Capybara's sake, whereas `no_label` means no identifier.

@Soullivaneuh on a side note, why does it say "1/12 tasks complete" under your referenced pull request? How do you enable tasks?

Soullivaneuh:
[...]

Issue #149, opened 2014-09-30
hide_label: :true doesn't work for checkboxes?


Qbd:
<%= f.check_box :attribute, hide_label: :true %> doesn't seem to work. Is there something I'm missing?

onebree:
@Qbd I am not sure if the gem has updated since your post, but try: `hide_label: true` instead. True can be typed as a boolean, not just a symbol.

Qbd:
Thanks for the help guys, I'm pretty sure I found a workaround but should I ever come across the problem again, I'll give more details.

Sorry, still kind of new at this whole open source thing

onebree:
@Qbd, please post your workaround, in case others have a similar issue. Alongside your current version of the gem :-)

Qbd:
I believe my workaround was to not use bootstrap forms..

I have a situation where a user can view tabular data and change certain boolean properties via AJAX submissions.

I just used the default form builder and declared f.checkbox.

At least looking back at my commits at that point in time, that seems to be the only major work I was doing with checkboxes.

abrambailey:
:(

CreativelyCloudy:
Hi all,

Had the same problem...found setting label: '' (empty quotes) did the trick.

chao-mu:
The above work arounds (specifying an empty label) do not work for me because they still render the label tag, throwing off my alignment. My work around was to just render a raw html checkbox and hope naming doesn't change. :-(.

My use case was checkboxes in a table.

chao-mu:
Another workaround could have been to "display: none" on the label with css.

PR #414, opened 2018-01-29
Update `hide_label` and `skip_label` for check boxes and radio buttons

lcreid:
This PR updates markup for check boxes and radio buttons to Bootstrap 4. It applies to the `radio` and `check_box` helpers only, not the `collection_radio_buttons` and `collection_check_boxes`. The collection versions should probably be discussed under #412.

Further clean-up of this code is possible, but an attempt was made to minimize changes (not always successful, mind you).

This is part of the response to #395.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 * Built-in support for the `nested_form` gem has been completely removed
 * The `icon` option is no longer supported (Bootstrap v4 does not include icons)
 * The deprecated Rails methods `check_boxes_collection` and `radio_buttons_collection` have been removed
+* `hide_label: true` and `skip_label: true` on individual check boxes and radio buttons apply Bootstrap 4 markup. This means the appearance of a page may change if you're upgrading from the Bootstrap 3 version of `bootstrap_form`, and you used `check_box` or `radio_button` with either of those options
 * Your contribution here!
 
 ### New features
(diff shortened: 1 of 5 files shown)


mattbrictson:
I think this looks great. The important part is that we have good test coverage for these new cases, which you have done. Thanks!

One question I have is whether `position-static` should be added for the `:hide_label` case. The Bootstrap docs don't seem clear on that. Does it look better one way or the other when rendered?

lcreid:
I convinced myself that `position-static` wasn't needed for the `hide_label` case. I believe I did one simple test on bootply.com, but I don't remember for sure. I'll try to find time to do a more careful test.

lcreid:
@mattbrictson good call on questioning me about `hide_label`! It *does* need `position_static`, as I hope [this](https://www.bootply.com/niyScCgLHt) shows. Without `position_static`, the buttons pile up on each other, and also on any other text or controls you have adjacent. So I've pushed an update and hopefully we can merge this soon. Again, thanks for catching that.

lcreid:
BTW, I'm pretty sure that `hide_label` and `skip_label` in legacy don't hide or skip the label on a check box or radio button, at least when specified to `check_box` and `radio_button`. I checked out legacy at one point, duplicated the basic check box test to two new tests, and added `hide_label: true` to one of the new tests, and `skip_label: true` to the other, and they both render labels with no apparent markup that would hide them.

It looks like this was discussed in #161. Someone suggested it wasn't needed because, at least for `skip_label`, you could use `check_box_without_bootstrap`. However, with the Bootstrap 4 markup, I think it would be nicer for the user if `skip_label` and `hide_label` both do what they're advertised to do.

I guess that means this might be considered a breaking change, since someone upgrading might get quite different markup compared to v2.7. At the very least, probably something for the "Upgrading" document.

It looks like this is the source of a number of issues: #162, #161, and #149.

Issue #413, opened 2018-01-29
Add Yard Documentation to Methods


lcreid:
I would like to start documenting some of the methods I'm working on, simply to help me keep track of what a method is supposed to do. It seems to me the best place to put this documentation would be in the source code so it can be automatically published by Yard when we publish the gem. 

This issue is to solicit feedback on the idea and to make sure I'm not going against the wishes of the community if I start documenting methods in the source.

I'm not proposing that we have to do a big push to document everything, although we could if someone wanted to. Nor am I proposing a particular technology. If there's a better alternative than Yard, let's use it.

GBH:
This is not even about Yard docs. It's about having helpful documentation so it's possible to figure out what a given method does. Methods like these: https://github.com/bootstrap-ruby/bootstrap_form/blob/master/lib/bootstrap_form/form_builder.rb#L106 have way too much stuff going on inside so it would be helpful to document all the possible options that can be passed in and what the resulting output will be. Splitting that method apart would help quite a bit too.

You just add documentation as you touch or figure out what methods actually do.

PR #420, opened 2018-01-31
Yard Documention First Attempt [skip ci]

lcreid:
This PR documents the form helpers, and a few of the field helpers. The documentation is a combination of what I understand the code says, and what I understand the documentation says. It's likely to be wrong in places on both counts.

This is a first response to issue #413. We can ignore this PR, heavily modify it, or merge it and work to improve it in the future.

There is a lot of redundancy. I'm not good enough at Yard to get rid of it. Help on that front is very welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ test/gemfiles/*.lock
 .ruby-version
 Vagrantfile
 .vagrant
+.yardoc
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,2 @@
+--markup-provider=redcarpet
+--markup=markdown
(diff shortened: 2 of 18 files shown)


GBH:
Oh wow that's overwhelming. Somehow I thought it would be something like https://github.com/rails/rails/blob/bdc581616b760d1e2be3795c6f0f3ab4b1e125a5/actionview/lib/action_view/helpers/form_helper.rb#L1345 But maybe because it's in rdoc format?

lcreid:
This PR has become too out of date, both in content and the organization of files in the source code.

PR #422, opened 2018-02-04
Move errors inside input-group and form-group.

### LINKED ISSUE #417: `invalid-feedback` should be inside `input-group`
The style that triggers display of the error is (amongst others) `.form-control.is-invalid ~ .invalid-feedback`.

Currently the `.invalid-feedback` div is outside the `.input-group` and as `.is-valid` is set on the input itself, it's never displayed.

This has been addressed in https://github.com/twbs/bootstrap/pull/25020

> Added support for form validation feedback in input groups. .input-group now has flex-wrap: wrap; on by default, allowing you to place validation feedback within the .input-group and style it based on the input's overall state. This isn't applicable to file inputs though.

This means the `.invalid-feedback` should be located as the last item inside `.input-group`.

I will take a look at this myself and see if I can fix, but it's probably an easy fix for someone who knows the code better. Thanks if you get there first!

### LINKED ISSUE #418: Errors on checkbox/radios collections should be inside the last element

lcreid:
It looks like Bootstrap 4 specifies that errors should be wrapped inside the input or form group (see https://github.com/twbs/bootstrap/pull/25020 and https://getbootstrap.com/docs/4.0/components/forms/#validation). This PR moves the error (and help) text inside the input-group and form-group.

(Based on the examples in the Bootstrap docs, the error and help text doesn't have to be inside an input or form group, but they do have to be within the same container as the `input` or `select`, and `label` (if any) of the control.)

The changes introduced by this PR can be summarized as:

- Since the majority of helpers called a method (`prepend_and_append_input`) to handle the `:append` and `:prepend` options, the error and help code was moved from `form_group` to `prepend_and_append_input`
- For the helpers that didn't call `prepend_and_append_input`, an "extract method" was done on `prepend_and_append_input` to separate out the code that makes the control, and optionally the error and help text (`control_error_help`). The helpers were modified to call `control_error_help`
- Help and error messages were added to `check_box` and `radio_button` methods, as they fit neither of the above models
- Tests were changed to move the help and error text inside the enclosing `div`

I couldn't see another way to move the error text inside the input group, without major changes to `form_group` and `form_group_builder`.

Because in effect the error handling is pushed down to the individual helpers, it's not longer possible to get an error message from `form_group` when the contents of the block that `form_group` calls don't include a call to a `bootstrap_form` helper. There is one test in `bootstrap_form_group_test.rb` that has a `form_group` around a `p` tag, and checks for the error message. This test has been commented out in this PR. 

NOTE: This PR contains the tests in PR #424, because the tests were needed to ensure this PR didn't introduce regressions.

Fixes #417.
Fixes #418.

### DIFF
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -77,16 +77,28 @@ module BootstrapForm
       end
 
       def prepend_and_append_input(name, options, error_text: nil, &block)
-        options = options.extract!(:prepend, :append, :input_group_class)
+        control_error_help(name,
+                           options,
+                           prepend: options.delete(:prepend),
+                           append: options.delete(:append),
+                           error_text: error_text,
+                           &block)
+      end
+
+      ##
+      # Render a block, and add its error or help.
+      # Add prepend and append if provided.
+      def control_error_help(name, options, prepend: nil, append: nil, error_text: nil, &block)
+        options = options.extract!(:input_group_class)
         input_group_class = ["input-group", options[:input_group_class]].compact.join(' ')
 
         input = capture(&block)
 
-        input = content_tag(:div, input_group_content(options[:prepend]), class: 'input-group-prepend') + input if options[:prepend]
-        input << content_tag(:div, input_group_content(options[:append]), class: 'input-group-append') if options[:append]
+        input = content_tag(:div, input_group_content(prepend), class: 'input-group-prepend') + input if prepend
+        input << content_tag(:div, input_group_content(append), class: 'input-group-append') if append
         input << error_text
         # FIXME: TBC The following isn't right yet. Wrap if there were errors. Maybe???
-        input = content_tag(:div, input, class: input_group_class) unless options.empty?
+        input = content_tag(:div, input, class: input_group_class) unless options.empty? && prepend.nil? && append.nil?
         input
       end
 
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -69,9 +69,11 @@ module BootstrapForm
       ##
       # Add prepend and append, if any, and error if any.
       # If anything is added, the whole thing is wrapped in an input-group.
+      # FIXME: I think this has to be the other way around, i.e. we need to be
+      # able to render fields with error or help, but only sometimes with
+      # prepend and append.
       def prepend_and_append_and_error_input(name, options, &block)
-        puts "options[:help]: #{options[:help]}"
-        prepend_and_append_input(name, options, error_text: generate_help(name, options[:help]).to_s, &block)
+        prepend_and_append_input(name, options, error_text: generate_help(name, options.delete(:help)).to_s, &block)
       end
 
       def prepend_and_append_input(name, options, error_text: nil, &block)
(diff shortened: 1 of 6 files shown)


test/bootstrap_form_group_test.rb
```diff
@@ -147,6 +147,32 @@ class BootstrapFormGroupTest < ActionView::TestCase
...
+            <div class="input-group-append">
+              <span class="input-group-text">.00</span>
+            </div>
+            <div class="invalid-feedback">can't be blank, is too short (minimum is 5 characters)</span>
+          </div>
+        </div>
+      </form>
+    HTML
+    # TODO: We should build the @builder properly from `bootstrap_form_for`, so it's easier to test errors.
+    assert_equivalent_xml expected, bootstrap_form_for(@user) { |f| f.text_field :email, prepend: '$', append: '.00' }
```
GBH: I would render `@builder.text_field :email, prepend: "foo", append: "bar"` instead of doing it though form_helper. Less html that way. All tests should just call helpers directly.

lcreid: That's how I started, but then it renders the default Rails error markup. `@builder` is missing the `bootstrap_form` override of the default Rails error rendering. I was going to enter an issue to change the way we make the builder, so it would render properly without the extra HTML, which I agree is a unnecessary and can be a pain.

GBH: You can put code used for error suppression (https://github.com/bootstrap-ruby/bootstrap_form/blob/master/lib/bootstrap_form/helper.rb#L45) into `setup` and `teardown` for tests that need it. You'll want to extract all tests that deal with error rendering into a separate file though.

lcreid: That's a good suggestion. In the spirit of limiting the scope of each PR, I continued to just tolerate a few extra lines of HTML in each test, but we should revisit the suggestion for easier error testing in another PR.

lcreid:
The latest push gives better test coverage, so the PR is ready for review.

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -66,15 +66,33 @@ def custom_control(*args, &block)
...
         input = capture(&block)
 
-        input = content_tag(:div, input_group_content(options[:prepend]), class: 'input-group-prepend') + input if options[:prepend]
-        input << content_tag(:div, input_group_content(options[:append]), class: 'input-group-append') if options[:append]
-        input = content_tag(:div, input, class: input_group_class) unless options.empty?
+        input = content_tag(:div, input_group_content(prepend), class: 'input-group-prepend') + input if prepend
+        input << content_tag(:div, input_group_content(append), class: 'input-group-append') if append
+        input << error_text
+        # FIXME: TBC The following isn't right yet. Wrap if there were errors. Maybe???
+        input = content_tag(:div, input, class: input_group_class) unless options.empty? && prepend.nil? && append.nil?
```
mattbrictson: Is this FIXME still valid? If so, could you clarify what additional cases need to be handled here?

mattbrictson:
Also, do `static_control` and `custom_control` lose their ability to display errors as a result of this PR?

lcreid:
Thanks for the feedback, @mattbrictson . I'm going to take another look at how this could be done without having to move the error reporting out of `form_group`. There are two challenges that I was running in to: 

- The way the radio buttons and check boxes are used, in some cases stand-alone and in some cases as a group inside a single form group
- The new concept of the input group, and the fact that the errors have to be wrapped inside the input group

Feel free to ask more questions or give suggestions. I'll make another branch locally and see if I can find a way to keep the error reporting behaviour more consistent with the old way, while still supporting the new requirements. If I figure something out, I'll push a different PR so you can see which makes more sense.

mattbrictson:
Closed in favor of #434

PR #424, opened 2018-02-05
Test the rendering of errors for various controls.

lcreid:
This pull request adds tests for various controls when there is an validation failure on the control. All these tests document the behaviour of the master branch as of the time the PR was created. 

Bootstrap 4's handling of errors is different that Bootstrap 3's, and has led to some issues already, e.g. #417 and #395. The tests here, therefore, document current functionality, so they give us a baseline for any required changes. These tests should not be considered a declaration about how `bootstrap_form` should render errors with Bootstrap 4. 

These tests also provide more coverage to ensure we don't introduce regressions as we address #417, #395 and any other issues that come up.

Given this PR contains only a few tests which don't significantly increase the run time of the test suite, I hope it's easy to merge. It would be great if someone other that @mattbrictson can review this PR, to take some of the load off Matt's shoulders.

### DIFF
--- a/test/bootstrap_fields_test.rb
+++ b/test/bootstrap_fields_test.rb
@@ -65,6 +65,21 @@ class BootstrapFieldsTest < ActionView::TestCase
     assert_equivalent_xml expected, @builder.file_field(:misc)
   end
 
+  test "file fields are wrapped correctly with error" do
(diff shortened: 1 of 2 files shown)


mattbrictson:
Nice work. Always good to have additional test coverage where it makes sense. 💯

PR #429, opened 2018-02-08
Add test to prove #427 is *not* an issue.

### LINKED ISSUE #427: Need `fields_with_bootstrap` method.
In preparing #369, it looked like there was no need to overload the Rails `fields` method with a `bootstrap_form` version of the helper. I now believe I may have been wrong about that. We need to test more to see whether the values of the following options default to the parent `form_with` options or not: `:control_col`, `:inline_errors`, `:label_col`, `:label_errors`, and `:layout`.

lcreid:
Add a test that proves we don't need a `fields_with_bootstrap` method. 

Fixes #427.


mattbrictson:
Makes sense 👍

Issue #432, opened 2018-02-16
Overriding `layout: :inline` at field level.


lcreid:
The [README](https://github.com/bootstrap-ruby/bootstrap_form#custom-field-layout) says that specifying a layout at the field level overrides the layout specified a the form level. The example shows a `:default` layout at the field level overriding a `:horizontal` form layout.

The documented example works, but it may be difficult to make the field-level `layout:` override a form-level `layout: :inline`. The form-level `class="form-inline"` appears to cause layout changes throughout the form that it's beyond my CSS abilities to figure out how to override.

I suggest we simply update the documentation to clarify that form-level `layout: :inline` can't be overridden, but I'm open to other suggestions.

mattbrictson:
> I suggest we simply update the documentation to clarify that form-level layout: :inline can't be overridden

Agreed 👍

PR #433, opened 2018-02-17
Update `layout: inline` at field level for Bootstrap 4.

### LINKED ISSUE #416: `collection_radio_buttons` and `collection_check_boxes` Need to be Reviewed for Bootstrap 4
The changes to `radio_with_bootstrap` and `check_box_with_bootstrap` may require some options that aren't currently being passed from their corresponding collection methods. This issue may require the following actions (feel free to suggest others):

- [ ] Write test cases to confirm the desired functionality. This may require a clearer definition of the functionality of the helpers
- [x] Review how the helpers functioned in the last released version. It's possible that the options weren't being passed before. If so, any changes would have to be reviewed in case they're breaking changes
- [ ] Implement the desired functionality

### LINKED ISSUE #426: `layout: :inline` at the field level doesn't render correctly for email (and probably other types)
Our test case for rendering a `layout: :inline` e-mail field within a `layout: :horizontal` form says the result should be:
```
<form accept-charset="UTF-8" action="/users" class="new_user" id="new_user" method="post" role="form">
  <input name="utf8" type="hidden" value="&#x2713;" />
  <div class="form-group">
    <label class="required" for="user_email">Email</label>
    <input class="form-control" id="user_email" name="user[email]" type="email" value="steve@example.com" />
  </div>
</form>
```
However, if you put that into Bootply, you can see that the label is rendered above the input field: https://www.bootply.com/Yk1F2jl2OL.

Adding `form-inline` to the `form-group` that wraps the controls seems to give better output, although there should probably be some spacing applied too. Perhaps we could look at how check boxes and radio buttons are formatted in-line.

This probably affects all input types except radio and checkbox, and likely selects as well.

Also, specifying `layout: :default` at the level of `form_group` or on a control may not cause the elements to be rendered the Bootstrap 4 default way, if the form was built with `layout: :inline`.

lcreid:
For `layout: :inline` elements to look right, Bootstrap 4 requires some markup at a "lower level" of the HTML than Bootstrap 3 did. This PR ensures that the elements correctly calculate whether the developer has requested in-line layout, and applies the appropriate classes.

In some cases, classes for in-line layout are applied regardless of whether the layout was specified at the form level or the field level. In other cases, the markup is only needed and applied if the in-line layout was only specified at the field level.

This PR adds a number of test cases for the six variations of form-level and field-level layout being different. The test cases test a few different types of fields:

- A text-like input, since a large number of types of input fields are handled internally like a text input
- A select, because selects are hendled differently
- Check box and radio button, since those two are handled very differently from the rest, and in fact are subtly different between the two
- A collection radio button, since it adds some additional handling not used by any of the above

I think it's important to have this level of test coverage, even if it could be argued that some of this might be tested elsewhere. (I thought that the date/time selects were similar enough to the others to forgo a test.)

There will be two outstanding issues after this PR is merged:

- If you have an in-line `form-group` with one or more radio buttons or check boxes inside it, the baseline of the `form-group`'s label won't be exactly the same as the baseline of the radio button or check box labels. (See https://www.bootply.com/fH5sF4fcju.)

- Overriding form-level `layout: :inline` at the field level does not work. See #432.

Fixes #426.
Fixes #416.


lcreid:
The tests aren't "real-world" in that the form layout is declared one way, and then all the fields inside are declared with another layout. However, our documentation says you can override the form level layout with field level layout, so I think it should be tested.

The test output that should look ugly in Bootply are where the form is `layout: :inline` and the fields are not. If you saw something unacceptable in any of the other combinations, please let me know.

Thanks for the the merges today, @mattbrictson ! I think we're getting close to alpha.

lcreid:
Doh! It took a while but I think I realized what you were getting at, @mattbrictson . I removed the tests that were testing the overrides of in-line output, so that in the future no one will think that, because it's in a test case, it's the correct output.

Since the links may not be anywhere else, here's what trying to override in-line layouts looks like at the time of writing: https://www.bootply.com/1hMYpBHds5 and: https://www.bootply.com/JqdpfgA36j.

Hopefully this can be approved now. :-)

mattbrictson:
Rebased

PR #435, opened 2018-02-20
Clarify overriding layouts [skip ci].

### LINKED ISSUE #432: Overriding `layout: :inline` at field level.

lcreid:
Bootstrap 4 makes it challenging to override form-level `layout: :inline`. This PR just documents that you can't do it.

@mattbrictson you agreed to this in https://github.com/bootstrap-ruby/bootstrap_form/issues/432#issuecomment-366471106, so approval should be easy.

Fixes #432.

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,7 +486,7 @@ The `label_col` and `control_col` css classes can also be changed per control:
 
 ### Custom Field Layout
 
-The `layout` can be overridden per field:
+The form-level `layout` can be overridden per field, unless the form-level layout was `inline`:
 
 ```erb
 <%= bootstrap_form_for(@user, layout: :horizontal) do |f| %>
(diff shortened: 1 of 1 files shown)


mattbrictson:
Looks good!

PR #436, opened 2018-02-20
Move validation errors for check and radio collections into last form-check div.

### LINKED ISSUE #418: Errors on checkbox/radios collections should be inside the last element

lcreid:
Bootstrap 4 validation errors for groups of check boxes or radio buttons are supposed to be inside the `<div class="form-group">` of the last check box or radio button in the group. See: https://getbootstrap.com/docs/4.0/components/forms/#supported-elements.

This PR contains some non-trivial refactoring of `check_box` and `radio_button`. Methods were extracted so that the `collection...` helpers could inject the error message when they knew they were working on the last control of the collection. The methods are private, but were left in the file near `check_box` and `radio_button` so it's hopefully a little easier to see the changes.

This PR also contains an instance variable that cause magic behaviour in `BootstrapForm::Helpers::Bootstrap#append_and_prepend_input`, so `form_group` doesn't output the error at the end of the form group that wraps the whole collection.

This PR was based off the code in PR #434 and depends on that code.

Fixes #418.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 * Built-in support for the `nested_form` gem has been completely removed
 * The `icon` option is no longer supported (Bootstrap v4 does not include icons)
 * The deprecated Rails methods `check_boxes_collection` and `radio_buttons_collection` have been removed
+* `hide_label: true` and `skip_label: true` on individual check boxes and radio buttons apply Bootstrap 4 markup. This means the appearance of a page may change if you're upgrading from the Bootstrap 3 version of `bootstrap_form`, and you used `check_box` or `radio_button` with either of those options
 * Your contribution here!
 
 ### New features
@@ -22,14 +23,15 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 * Support for Rails 5.1 `form_with` - [@lcreid](https://github.com/lcreid).
 * Support Bootstrap v4's [Custom Checkboxes and Radios](https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios-1) with a new `custom: true` option
 * Allow HTML in help translations by using the `_html` suffix on the key - [@unikitty37](https://github.com/unikitty37)
+* [#408](https://github.com/bootstrap-ruby/bootstrap_form/pull/408): Add option[:id] on static control #245 - [@duleorlovic](https://github.com/duleorlovic).
 * Your contribution here!
 
 ### Bugfixes
 
-* Your contribution here!
 * [#357](https://github.com/bootstrap-ruby/bootstrap_form/pull/357) if provided,
   use html option `id` to specify `for` attribute on label
   [@duleorlovic](https://github.com/duleorlovic)
+* Your contribution here!
 
 ## [2.7.0][] (2017-04-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 ### Bugfixes
 
 * Your contribution here!
+* [#357](https://github.com/bootstrap-ruby/bootstrap_form/pull/357) if provided,
+  use html option `id` to specify `for` attribute on label
+  [@duleorlovic](https://github.com/duleorlovic)
 
 ## [2.7.0][] (2017-04-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 
 ### New features
 
+* Support for Rails 5.1 `form_with` - [@lcreid](https://github.com/lcreid).
 * Support Bootstrap v4's [Custom Checkboxes and Radios](https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios-1) with a new `custom: true` option
 * Allow HTML in help translations by using the `_html` suffix on the key - [@unikitty37](https://github.com/unikitty37)
 * Your contribution here!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 
 ### Breaking changes
 
+* Rails 4.x is no longer supported
+* Ruby 2.2 or newer is required
 * Built-in support for the `nested_form` gem has been completely removed
 * The `icon` option is no longer supported (Bootstrap v4 does not include icons)
 * The deprecated Rails methods `check_boxes_collection` and `radio_buttons_collection` have been removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 ## [2.7.0][] (2017-04-21)
 
 Features:
-  * [#325](https://github.com/bootstrap-ruby/rails-bootstrap-forms/pull/325): Support :prepend and :append for the `select` helper - [@donv](https://github.com/donv).
+  * [#325](https://github.com/bootstrap-ruby/bootstrap_form/pull/325): Support :prepend and :append for the `select` helper - [@donv](https://github.com/donv).
 
 ## [2.6.0][] (2017-02-03)
 
@@ -37,7 +37,7 @@ Bugfixes:
   - Fix ambiguous first argument warning (#311, @mikenicklas)
 
 Features:
-  - Add a FormBuilder#custom_control helper [#289](https://github.com/bootstrap-ruby/rails-bootstrap-forms/pull/289)
+  - Add a FormBuilder#custom_control helper [#289](https://github.com/bootstrap-ruby/bootstrap_form/pull/289)
 
 ## [2.5.3][] (2016-12-23)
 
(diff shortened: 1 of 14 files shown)


lcreid:
There are some hard problems here. I can't see how we can output Bootstrap 4 markup without requiring some users to change their code. I hope someone can guide me to what we should do. Sorry this is so convoluted, but I can't see a way to separate one issue from another. @mattbrictson @desheikh @donv any thoughts?

## Background
### Bootstrap 3 (v2.7)
There are basically four use cases we're interested in:

1. Rendering any control except a bare check box or radio button outputs a validation error message and styles the control appropriately, if the object has a validation error. There are no tests in branch `legacy-2.7` for the default case, but `test/bootstrap_form_test.rb` line 71 (and others) test default error messages and label error messages. Rendered output: https://www.bootply.com/AoHBiFn0wg
2. Rendering one or more check boxes or radio buttons inside a block given to `form_group` outputs a single validation error message after the check boxes or radio buttons, and styles the control appropriately, if the object has a validation error. There are no tests in branch `legacy-2.7` for this. Rendered output: https://www.bootply.com/gLvYyaA6zr and https://www.bootply.com/tUxPJEklIJ
3. Rendering a bare check box or radio button not from a block given to `form_group` does not output an error message or style the control appropriately, if the object has a validation error. There are no tests in branch `legacy-2.7` for this. 
4. Rendering any other text from a block given to `form_group` outputs a validation error message and styles the contents of the block appropriately, if the first argument to `form_group` is an object with a validation error (In branch `legacy-2.7`, `test/bootstrap_form_group_test.rb` at line 171). It renders with the error message in red, but the "field" is not in red. Rendered output: https://www.bootply.com/v4HtoM93EE.

### Bootstrap 4
The markup is different. The change that's causing grief is that check boxes and radio buttons are now rendered inside a `div.form-check`, and the validation error message for a list of check boxes or radio has to be *inside* the *last* check box's or radio button's `div.form-check`. That affects the above use cases in the following ways:

1. All helpers except `check_box`, `collection_check_boxes`, `collection_radio_buttons`, and `radio_button` work as of PR #434 (although note issue #444). Bootstrap 4 added a `div.input-group` for append and prepend which required some non-trivial modifications, but the overall structure of our code is pretty similar to v2.7
2. As mentioned above, each check box and radio button now has to be wrapped in a `div.form-check`. Error and help messages have to be inside the last `div.form-check`. This means we can no longer simply add error messages after rendering one or more check boxes or radio buttons
3. Ideally, whatever solution is chosen for 4.0 does *not* output error messages from all instances of a bare `check_box` or `radio_button`, since that's not what happened in v2.7
4. For this use case at the time of writing, even though we output the validation error, it won't be displayed because of the way the Bootstrap 4 classes work. See: https://www.bootply.com/nv1z9gbUKv for an example of the mark-up. The Bootstrap 4 mixin for `.is-invalid` starts here: https://github.com/twbs/bootstrap/blob/2f4e785a8c623fbe1b011edabecafe99b9f0817b/scss/mixins/_forms.scss#L29, and the CSS that causes the display of `.invalid-feedback` is here: https://github.com/twbs/bootstrap/blob/2f4e785a8c623fbe1b011edabecafe99b9f0817b/scss/mixins/_forms.scss#L53-L136. This is because `.invalid-feedback` includes a `display: none`, which is overridden by `display: block` depending on the validation state of certain siblings. Unless the programmer ensures that the text in the block is marked up just the right way, the validation message won't be visible anyway

## Possible Solutions
These are not in any particular order of preference, effort, or any other criteria.

### A
- Change `form_group` to parse the HTML returned from its block, determine if the last direct child is a radio button or check box (`div.form-check`), and add the error inside that child. Otherwise, add the error at the end as always. This still doesn't make `form_group`-wrapped arbitrary text look exactly like legacy, because of use case 4 in the "Bootstrap 4" section above.
- Document how to mark up the text from the block so that errors will be displayed correctly (although this slightly changes the layout of the text)
- and/or
- Encourage the use of `static_control` if all the programmer is doing is trying to get a static or read-only control

This is the only option I can think of that avoids changing `check_box`, `radio_button`, `collection_check_boxes`, and `collection_radio_button`. Nokogiri is already loaded in Rails applications. Still, it feels like a heavy-weight solution to me.

### B

- Add an `error_message: true` option to `check_box` and `radio_button` to cause them to output the object's validation error messages. The default is `false` to preserve the current behaviour that `check_box` and `radio_button` don't output error messages
- Move stuff around internally so that `form_group` no longer outputs error messages, but rather it's pushed to the appropriate place so that `check_box` and `radio_button` output the errors. PR #422 actually did this, but we decided at the time that we didn't want to break the "arbitrary text in the `form_group` block" case
- Document that the last `check_box` or `radio_button` in a `form_group` has to be called with `error_message: true` in order to render the validation error message
- Document that arbitrary text in a `.form-group` should be coded using raw Bootstrap, rather than by using the `form_group` helper

### C

- New helper `check_form_group` for wrapping check boxes and radio buttons only
- Add an `error_message: true` option to `check_box` and `radio_button` to cause them to output the object's validation error messages
- Document that the last `check_box` or `radio_button` in a `check_form_group` has to be called with `error_message: true` in order to render the validation error message

This solution is easy to document, i.e. the programmer just has to change all `form_group`s wrapping check boxes or radio buttons, to `check_form_group`. The programmer also has to change the last helper in a group to have the `error_message: true` option.

Internally, it's likely that `form_group` and `check_form_group` can share almost all their code, although it might require some non-trivial refactoring to do so elegantly.

### D

- Add a helper to output non-control blocks (`raw_form_group`?), i.e. anything *not* wrapping check boxes or radio buttons
- Change `form_group` to not output validation errors, and push that down to the relevant helpers
- Add an `error_message: true` option to `check_box` and `radio_button` to cause them to output the object's validation error messages
- Document that the last `check_box` or `radio_button` in a `form_group` has to be called with `error_message: true` in order to render the validation error message

This is pretty much the same as Solution C, except it requires the code changes for the "arbitrary text" use case, which I think might be the less common one compared to the "check box or radio button" use case.

mattbrictson:
Thanks for digging into this one and documenting all the approaches.

To be honest, none of the solutions feels right to me. Solution "A" is the most desirable end-user experience, but I think your intuition is correct that this is too heavy-handed.

This might be a crazy idea, but can we solve this with just CSS? In other words, can we add our own styles that make the error messages appear even if we render the HTML the "wrong" way? Or perhaps a JS snippet that manipulates the DOM?

Also: do we know if/how simple_form has dealt with this problem? Maybe that project has some ideas we can borrow.

lcreid:
@mattbrictson your timing is perfect. I was just sitting down to charge off with Solution B, but the CSS idea is worth pursuing. I will look into it today. I'll also look at `simple_form`.

Just a note that I've spent some time looking at the Bootstrap 4 source. I think it confirms that some of the struggles are because Bootstrap 4 works in a fundamentally different way, and we may want to consider adopting the "Bootstrap 4 Way", at the expense of requiring more from programmers migrating from `bootstrap_form` 2.7 to `bootstrap_form` 4.0.

lcreid:
`simple_form` when using Bootstrap 4 seems to output error messages as if they were help text. Error messages don't get different colours. Also, it looks as if some of the markup is still Bootstrap 3-style.

It looks like errors from check boxes and radio buttons in `simple_form` don't have to deal with the same issues as we do, partly because they're not trying to do things the "Bootstrap 4" way, and partly because the easy way to generate a group of check boxes or radio buttons is with a collection, which they can then handle where to put the error message.

I should spend more time looking at `simple_form`, but for this issue, I don't think there's much more to learn from `simple_form`.

I like the fact that `bootstrap_form` produces output that looks just like the Bootstrap documentation says it should look.

lcreid:
About use case 4 above, the case of a `form_group` with a block that returns arbitrary HTML:

One possible solution is simply to tell the programmer to add the following to their (S)CSS:
```
.invalid-feedback { display: block; }
```
This is a rather large hammer to take to the problem, but given the way `bootstrap_form` works (and worked in the past), I think it's relatively safe to recommend as a solution. I'm specifically suggesting that we document this in the migration guide, and leave it up to the programmer to decide whether they need to add the CSS. It should only be needed when rendering HTML without a `bootstrap_form`-generated input tag in the block given to `form_group`. Otherwise, we've already made `bootstrap_form` work with Bootstrap 4 errors, except for radio buttons and check boxes.

lcreid:
Finally, for the issue of how to handle error messages when there's more than one `check_box` or `radio_button` in a block to `form_group`: I have an idea for how to do it. It was half-formed in my head before, but looking at the `simple_form` code seems to have helped crystallize it. But I need to actually try some code and see if the idea will work. I'll let you know in a day or two.

Spoiler alert: My idea will require creating some new classes, not visible to the end programmer, and likely moving some of the code, specifically the code in `check_box` and `radio_button` into those classes. 

@mattbrictson if you have any thoughts about this or my previous comments today, I'm all ears. Or you can wait to see what I come up with.

mattbrictson:
OK, I will wait to see your solution. In the meantime: I agree that it is important for bootstrap_form to output correct Bootstrap v4 markup, and if we require a little extra from the programmer in edge cases, then that is an acceptable compromise. We don't want to rely on "magic".

I'm also on board with documenting the `display: block` hammer for those rare (?) cases that need it.

lcreid:
The new idea I had [yesterday](https://github.com/bootstrap-ruby/bootstrap_form/pull/436#issuecomment-370973083) turned out to require more `ActionView`-related knowledge that I felt comfortable developing in the short term, and I could think of cases where v4.0 would still not be 100% non-breaking with v2.7. So I've basically implemented [Solution B above](https://github.com/bootstrap-ruby/bootstrap_form/pull/436#issuecomment-369350477). I want to take one more look at it tonight, and then I'll push a new PR, so we can compare and choose the best approach.

lcreid:
This PR is no longer need since PR #448 has been merged.

PR #306, opened 2016-11-10
Bootstrap v4 file field

trammel:
Changed the file input helper to match Bootstrap v4's new expected HTML as described http://v4-alpha.getbootstrap.com/components/forms/#file-browser

Also, added support for passing attributes to the span section, so JS can be used to update the prompt text into a filename, as described https://github.com/twbs/bootstrap/issues/19661

PR #439, opened 2018-02-25
Bootstrap v4 file input

mattbrictson:
This is a rebase and redo of #306 by @trammel.

I did the minimum necessary (I think) to generate the appropriate markup [explained in the v4 docs](http://getbootstrap.com/docs/4.0/components/forms/#file-browser).

```html
<div class="form-group">
  <label for="user_misc">Misc</label>
  <div class="custom-file">
    <input class="custom-file-input" id="user_misc" name="user[misc]" type="file" />
    <label class="custom-file-label" for="user_misc">Choose file</label>
  </div>
</div>
```

Note that in Bootstrap v4:

> We hide the default file `<input>` via opacity and instead style the `<label>`.

In other words, the `label.custom-file-label` is styled to look like a text input, and the contents of that label are used for the placeholder text inside that pseudo-text input. The placeholder is "Choose file" by default and can be customized using the `:placeholder` option, if desired.

There might be some edge cases I missed. But, good enough for alpha?

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -61,9 +61,18 @@ module BootstrapForm
 
     def file_field_with_bootstrap(name, options = {})
       prevent_prepend_and_append!(options)
-      options = options.reverse_merge(control_class: 'form-control-file')
+      options = options.reverse_merge(control_class: "custom-file-input")
       form_group_builder(name, options) do
-        file_field_without_bootstrap(name, options)
+        content_tag(:div, :class => "custom-file") do
+          placeholder = options.delete(:placeholder) || "Choose file"
+          placeholder_opts = { class: "custom-file-label" }
+          placeholder_opts[:for] = options[:id] if acts_like_form_tag
+
+          input = file_field_without_bootstrap(name, options)
+          placeholder_label = label(name, placeholder, placeholder_opts)
+          concat(input)
+          concat(placeholder_label)
+        end
       end
     end
 
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -63,10 +63,10 @@ module BootstrapForm
       prevent_prepend_and_append!(options)
       options = options.reverse_merge(control_class: "custom-file-input")
       form_group_builder(name, options) do
-        content_tag(:div, :class => "custom-file") do
+        content_tag(:div, class: "custom-file") do
           placeholder = options.delete(:placeholder) || "Choose file"
           placeholder_opts = { class: "custom-file-label" }
-          placeholder_opts[:for] = options[:id] if acts_like_form_tag
+          placeholder_opts[:for] = options[:id] if options[:id].present?
 
           input = file_field_without_bootstrap(name, options)
           placeholder_label = label(name, placeholder, placeholder_opts)
--- a/test/bootstrap_fields_test.rb
+++ b/test/bootstrap_fields_test.rb
@@ -59,7 +59,10 @@ class BootstrapFieldsTest < ActionView::TestCase
     expected = <<-HTML.strip_heredoc
       <div class="form-group">
         <label for="user_misc">Misc</label>
-        <input class="form-control-file" id="user_misc" name="user[misc]" type="file" />
+        <div class="custom-file">
+          <input class="custom-file-input" id="user_misc" name="user[misc]" type="file" />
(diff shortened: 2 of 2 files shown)


mattbrictson:
Here's how it renders in the demo app:

screen shot 2018-02-24 at 4 30 24 pm

lib/bootstrap_form/form_builder.rb
```diff
@@ -61,9 +61,18 @@ def initialize(object_name, object, template, options)
...
     def file_field_with_bootstrap(name, options = {})
       prevent_prepend_and_append!(options)
-      options = options.reverse_merge(control_class: 'form-control-file')
+      options = options.reverse_merge(control_class: "custom-file-input")
       form_group_builder(name, options) do
-        file_field_without_bootstrap(name, options)
+        content_tag(:div, :class => "custom-file") do
+          placeholder = options.delete(:placeholder) || "Choose file"
+          placeholder_opts = { class: "custom-file-label" }
+          placeholder_opts[:for] = options[:id] if acts_like_form_tag
```
lcreid: For Rails 5.1, we may need to copy options[:id] to placeholder_opts[:for] in all cases, not just if `acts_like_form_tag`, in case they're using `bootstrap_form_with`, and therefore the helpers aren't generating IDs.

lib/bootstrap_form/form_builder.rb
```diff
@@ -61,9 +61,18 @@ def initialize(object_name, object, template, options)
 
     def file_field_with_bootstrap(name, options = {})
       prevent_prepend_and_append!(options)
-      options = options.reverse_merge(control_class: 'form-control-file')
+      options = options.reverse_merge(control_class: "custom-file-input")
       form_group_builder(name, options) do
-        file_field_without_bootstrap(name, options)
+        content_tag(:div, :class => "custom-file") do
```
lcreid: So we allow the programmer to use only the Bootstrap 4 custom file dialog, and not the browser's built-in dialog?

test/bootstrap_fields_test.rb
```diff
@@ -72,7 +75,10 @@ class BootstrapFieldsTest < ActionView::TestCase
...
       <div class="form-group">
         <label for="user_misc">Misc</label>
-        <input class="form-control-file is-invalid" id="user_misc" name="user[misc]" type="file" />
+        <div class="custom-file">
+          <input class="custom-file-input is-invalid" id="user_misc" name="user[misc]" type="file" />
+          <label class="custom-file-label" for="user_misc">Choose file</label>
+        </div>
         <div class="invalid-feedback">error for test</div>
       </div>
     </form>
```
lcreid: I would add a `form_with` test under Rails 5.1 passing in a explicit ID, and making sure the ID is on the `input` and used as the `for` attribute of the `label`.

lib/bootstrap_form/form_builder.rb
mattbrictson: To clarify: the "custom" part is how the input is styled. When the user clicks on the "browse" button, they still get the browser's built-in dialog. So I think this is OK.

test/bootstrap_fields_test.rb
mattbrictson: Good suggestion. I will give that a shot!

mattbrictson:
@lcreid I added a couple more commits to handle the `form_with` scenario and add some more tests. Good catch!

test/bootstrap_fields_test.rb
```diff
@@ -59,20 +59,54 @@ class BootstrapFieldsTest < ActionView::TestCase
...
+        <div class="custom-file">
+          <input class="custom-file-input" id="user_misc" name="user[misc]" type="file" />
+          <label class="custom-file-label" for="user_misc">Pick a file</label>
+        </div>
+      </div>
+    HTML
+    assert_equivalent_xml expected, @builder.file_field(:misc, placeholder: "Pick a file")
+  end
+
+  if ::Rails::VERSION::STRING > '5.1' && ::Rails::VERSION::STRING < '5.2'
```
lcreid: This looks good. If I really wanted to nit-pick, I would suggest changing the ID you use to be something other than what Rails != 5.1 would use, and then the test would be valid for all Rails versions >= 5.1. It would give coverage for that code path, since this helper doesn't use the standard `form_group_builder`/`form_group` approach.

mattbrictson: ✅ done

Issue #444, opened 2018-02-28
Help text rendered inside input-group instead of form-group


ehutzelman:
I believe this may be related to #434 

When using help text along with a prepend, the following structure is generated with the help text being rendered inside the input-group:
```html
<div class="form-group">
  <label>Label</label>
  <div class="input-group">
    <div class="input-group-prepend">
      <span class="input-group-text">$</span>
    </div>
    <input class="form-control" type="text" name="my_field">
    <small class="form-text text-muted">help text here...</small> <!-- Original location -->
  </div>
</div>
```
This is causing rending issues at some widths with the help text being rendered to the right of the input field instead of below it. When moving the help text outside of the input-group and inside the outer form-group, the issue seems to be fixed and the help text renders underneath the input field as expected:
```html
<div class="form-group">
  <label>Label</label>
  <div class="input-group">
    <div class="input-group-prepend">
      <span class="input-group-text">$</span>
    </div>
    <input class="form-control" type="text" name="my_field">
  </div>
  <small class="form-text text-muted">help text here...</small>  <!-- Moved here -->
</div>
```

lcreid:
Thanks for taking the time to report this. This is is a great bug report. 

In v2.7 of `bootstrap_form`, aka the Bootstrap 3 version, the help text was not rendered if there was a validation error. No one currently active on `bootstrap_form` knows why for sure, although https://github.com/bootstrap-ruby/bootstrap_form/issues/284#issuecomment-362791090 gives a plausible explanation.

Whatever the reason, the code for help and error messages is basically in one method. We've been trying to minimize code changes as much as practical for the Bootstrap 4 version, so we rendered help and errors inside `.input-group` with the same method. This means they come out in the same place. #417 noted that the validation errors need to be inside the `.input-group`. Your report confirms that the help text has to be outside the `.input-group`.  So something has to change. :-)

@ehutzelman would you like to take a shot at a PR for the `.input-group` case? If so, please comment here and/or assign the issue to yourself. If not, no worries. We like to encourage the community to contribute, but if you can't, we'll fix this sooner rather than later.

ehutzelman:
Thanks for the quick response @lcreid. I've overbooked myself for March, but I could possibly pick this up in April if it remains unresolved. If anyone gets to it before then, more power to them :)

lcreid:
[...]

PR #445, opened 2018-03-02
Issue #444 - Help text rendered in wrong place.

parkeryoung:
The help text and error message are no longer rendered in the same
place. The help text is now rendered in the `form-group` div instead
of in the `input-group` div. Before we could render the error and help
text together, however that has changed and they need to be separate.

With this separation we need to check in the `generate_help` method to
see if there are errors and if inline errors are being rendered; as we
don't want to render the help text then.

Note - 
At first it seemed a little strange that the help text is not being rendered in the event of an error. It does seem to make sense and look better if they aren't both rendered at the same time. I don't know if I'd want to render them both at the same time ever, but I'm curious of others might. They could make the failing validation message say the same thing that the help text said. This project looks like it could go option crazy if the maintainers aren't careful so it might be best to not have this. Just a thought I had while working on this.

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -263,7 +263,11 @@ module BootstrapForm
           control = content_tag(:div, control, class: control_class)
         end
 
-        concat(label).concat(control)
+        help = options[:help]
+
+        help_text = generate_help(name, help).to_s
+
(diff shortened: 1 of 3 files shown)


mattbrictson:
This looks great, thanks! 🙇‍♂️

PR #446, opened 2018-03-04
Fixes #443 - Updates `static_control` method.

### LINKED ISSUE #443: `static_control` needs to be updated to Bootstrap 4.
Static controls are currently implemented as a `p` tag that contains the text of the field, styled with `.form-control-static`. That class no longer exists in Bootstrap 4. It appears that `.form-control-static` was intended to render static text vertically aligned with form labels, in horizontal layout and in-line layout forms.

In Bootstrap 4 there is a `.form-control-plaintext`, but it appears to be intended to be applied to a read-only input control (https://getbootstrap.com/docs/4.0/components/forms/#readonly-plain-text).

Simply applying `.form-control-plaintext` to text inside a `form_group` block will not render correctly if there's a validation error. In Bootstrap 4, the error highlight activates on `form-control.is-invalid`. The `p` tag has to have `.form-control.form-control-plaintext.is-invalid` for the invalid feedback to appear, and for the styling (red outline) to happen on the paragraph text. 

Further, if the intent of `static_control` is to show something that looks like an input field, but that the user can't change, why wouldn't the developer simply code the field with the `readonly` attribute, like [here](https://www.bootply.com/4rds6D8osj), or as a disabled input field?

Apologies if the above doesn't make sense. I've been staring at this too long, and I can't make it any better.

Would it make sense to do this:

- Change `static_control` to render a text input field with `.form-control-static` added to it
- Document that programmers should just use disabled or readonly inputs, with `.form-control-plaintext` if they don't want the border around the field
- If there isn't much use for `static_control`, deprecate it.

parkeryoung:
In Bootstrap 4 the `form-control-static` class no longer exists. It
has been replaced with using the `form-control-plaintext` class on a
readonly input. This reflects that change in approach.

There are a couple of potential problems/oddities.

From a backwards-compatabilty standpoint, if people have been putting
other tags inside of the block this will likely break whatever they
may have been trying to do. There doesn't seem to be any way around
this limitation as it is caused by the text input approach.

Secondly, this approach seems to break convention in this project a
little. It doesn't seem as if any of the `_with_bootstrap` methods are
directly referenced anywhere. I'm not sure how much of a problem this
is, it seemed like the best way to stay DRY.

Should I also update the CHANGELOG? I don't know if it was an accident or not with the last contribution, but nothing was added to it.

### DIFF
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -48,15 +48,20 @@ module BootstrapForm
         options = args.extract_options!
         name = args.first
 
-        html = if block_given?
+        value = if block_given?
           capture(&block)
         else
           object.send(name)
         end
 
-        form_group_builder(name, options) do
-          content_tag(:p, html, { class: static_class }.merge(options[:id].present? ? { id: options[:id] } : {}))
-        end
+        static_options = {
+          value: value,
+          class: static_class,
+          readonly: true,
+          control_class: nil
+        }
+
+        text_field_with_bootstrap(name, options.merge(static_options))
       end
 
       def custom_control(*args, &block)
@@ -94,7 +99,7 @@ module BootstrapForm
       end
 
       def static_class
-        "form-control-static"
+        "form-control-plaintext"
       end
     end
   end
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -44,25 +44,18 @@ module BootstrapForm
         end
       end
 
-      def static_control(*args, &block)
+      def static_control(*args)
         options = args.extract_options!
         name = args.first
 
-        value = if block_given?
-          capture(&block)
-        elsif options[:value]
-          options[:value]
-        else
-          object.send(name)
-        end
-
-        static_options = {
-          value: value,
+        static_options = options.merge({
           readonly: true,
-          control_class: static_class
-        }
+          control_class: [options[:control_class], static_class].compact.join(" ")
+        })
 
-        text_field_with_bootstrap(name, options.merge(static_options))
+        static_options[:value] = object.send(name) if static_options[:value].nil?
+
+        text_field_with_bootstrap(name, static_options)
       end
 
       def custom_control(*args, &block)
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -50,15 +50,16 @@ module BootstrapForm
 
         value = if block_given?
           capture(&block)
+        elsif options[:value]
+          options[:value]
         else
           object.send(name)
         end
 
         static_options = {
           value: value,
-          class: static_class,
           readonly: true,
-          control_class: nil
+          control_class: static_class
         }
 
         text_field_with_bootstrap(name, options.merge(static_options))
--- a/test/bootstrap_form_group_test.rb
+++ b/test/bootstrap_form_group_test.rb
@@ -237,14 +237,14 @@ class BootstrapFormGroupTest < ActionView::TestCase
 
   test "form_group creates a valid structure and allows arbitrary html to be added via a block" do
     output = @horizontal_builder.form_group :nil, label: { text: 'Foo' } do
-      %{<p class="form-control-static">Bar</p>}.html_safe
+      %{<input class="form-control-plaintext" value="Bar">}.html_safe
     end
 
     expected = <<-HTML.strip_heredoc
       <div class="form-group row">
         <label class="col-form-label col-sm-2" for="user_nil">Foo</label>
         <div class="col-sm-10">
-          <p class="form-control-static">Bar</p>
+          <input class="form-control-plaintext" value="Bar">
         </div>
       </div>
     HTML
@@ -253,13 +253,13 @@ class BootstrapFormGroupTest < ActionView::TestCase
 
   test "form_group adds a spacer when no label exists for a horizontal form" do
     output = @horizontal_builder.form_group do
-      %{<p class="form-control-static">Bar</p>}.html_safe
+      %{<input class="form-control-plaintext" value="Bar">}.html_safe
     end
 
     expected = <<-HTML.strip_heredoc
       <div class="form-group row">
         <div class="col-sm-10 offset-sm-2">
-          <p class="form-control-static">Bar</p>
+          <input class="form-control-plaintext" value="Bar">
         </div>
       </div>
     HTML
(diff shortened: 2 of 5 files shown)


lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -48,15 +48,20 @@ def static_control(*args, &block)
...
         end
 
-        form_group_builder(name, options) do
-          content_tag(:p, html, { class: static_class }.merge(options[:id].present? ? { id: options[:id] } : {}))
-        end
+        static_options = {
+          value: value,
+          class: static_class,
+          readonly: true,
+          control_class: nil
```
lcreid: I'm curious why you assign `nil` to `control_class` here, and pass `static_class` via the `:class` key (two lines above)?

test/bootstrap_form_group_test.rb
```diff
@@ -237,14 +237,14 @@ class BootstrapFormGroupTest < ActionView::TestCase
 
   test "form_group creates a valid structure and allows arbitrary html to be added via a block" do
     output = @horizontal_builder.form_group :nil, label: { text: 'Foo' } do
-      %{<p class="form-control-static">Bar</p>}.html_safe
+      %{<input class="form-control-plaintext" value="Bar">}.html_safe
```
lcreid: We like to keep PRs very focused if possible. In this case we should only be changing the test cases that actually include a call to `static_control`. Sorry if my ruminations in the issue confused things. Could you please revert this file to the existing tests? Thanks in advance.

lib/bootstrap_form/helpers/bootstrap.rb
parkeryoung: Thanks for pointing this out, it was just an artifact of me walking through the problem and something I didn't realize wasn't necessary later.

A little after this as well these options are merged into the options that are passed into this method, so they will be overwritten. I think that with `control_class` that should be okay because we have the `class` option. The `readonly` option should be okay as well as it is necessary for the functionality. 

The only one I'm a bit iffy about is the `value` option. Now this can be set in the block, but I think it would probably be best to make it consistent with some of the other methods as well. I'll go ahead and make that change as well. If you have any objections let me know.

test/bootstrap_form_group_test.rb
parkeryoung: Okay, I made this change because the `form-control-static` class no longer exists and I was trying to purge the project of any reference to it. We could do it in a separate PR, but it seemed related so I made the change in this one.

parkeryoung:
I don't think there's anything in this that will make the error not show up for the `static_control`. I haven't tested errors and don't know what it will look like. So if they do not work together, it seems that will be a problem with bootstrap.

If it does not work properly, perhaps we do want to muffle these errors, however that could prove to be frustrating if we deliberately remove error messages. I don't think that's something we'd want.

lib/bootstrap_form/helpers/bootstrap.rb
lcreid: I'm not sure I follow you completely. 

I think what you're doing with `value` is pretty good. Although it might be interesting to only assign to `options[:value]` if the block is given. That way if the user explicitly gives a `:value` option, it will work as they might expect. I think something like this:
```
options[:value] = yield if block_given?
```
I guess I would have thought it would be easiest to just add `form-control-plaintext` to anything that already happened to be in `options[:control_class]` and then let the logic of `text_field` do the rest. Something like:
```
options[:control_class] = [options[:control_class], static_class].compact.join(" ")
```
Make sure you check what I typed above, and that it makes sense. I hope you don't mind the nit-picking.

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -48,15 +48,21 @@ def static_control(*args, &block)
...
           object.send(name)
         end
 
-        form_group_builder(name, options) do
-          content_tag(:p, html, { class: static_class }.merge(options[:id].present? ? { id: options[:id] } : {}))
-        end
+        static_options = {
+          value: value,
+          readonly: true,
+          control_class: static_class
```
lcreid: If the user passed in an `option[:control_class]`, we want to preserve it and add `static_class` to the list of `control_class`es. I believe the way this is merged below will overwrite whatever the user passes in for `option[:control_class]`

mattbrictson:
This looks good to me. Since `static_control` now just delegates to `text_field_with_bootstrap`, I expect that error messages, etc. would continue to work as with other Bootstrap fields. Anyway, I think showing an error on a read-only field is an odd case that we don't need to worry too much about.

I appreciate that the PR attempts to maintain backwards compatibility by continuing to accept a block for `static_control`, but I think the usage of that block is different enough that it is effectively still a breaking change. Before, the block could be arbitrary HTML and would be nested within a `<p>`, whereas now the block is interpreted as a field value to be placed in the `value` attribute of an `<input>`.

My suggestion would be to remove support for the block and explain this change in the CHANGELOG.

Otherwise I'd say this PR is good to go. Thanks for the contribution!

test/bootstrap_form_group_test.rb
mattbrictson: I'm OK with leaving these changes in. I like the idea of removing all references to `form-control-static` in this PR.

lib/bootstrap_form/helpers/bootstrap.rb
parkeryoung: Can do, is `control_class` specifically put on the `input`(or whatever form element) and the `class` option put on the surrounding div? I'm not entirely certain what the difference between these two options is.

parkeryoung:
@mattbrictson perhaps getting rid of the block is a better way to go about it as it will make it fail hard and will likely be easier to find in automated tests instead of silently changing the functionality. I am curious how people actually used that block and why they would put nested html inside of it, but I'm fine with removing it entirely.

lib/bootstrap_form/helpers/bootstrap.rb
lcreid: According to the documentation, the `:control_class` string is added to the classes that are put on the `input` tag. The `:wrapper_class` string is added to the classes that are put on the `div.form-group` wrapper. `:class`.

It's very hard to tell what exactly is going on in that code. Even though I've spent the last two months off and on staring at that code, I'm going to have to take some time be sure. Sorry I can't be more help right now. I'll try to take a look at this tonight (Pacific Time).

It might be worth your time to find some test cases that test those options, and see what they do. Maybe add a few test cases if you're still in doubt about what the code does.

lcreid: OK. My bad. Sorry if I misled you. `:control_class` will not apply the `.form-control` that we need. However, if you set `options[:class] = static_class` it works (as I think you figured out a day or two ago). My apologies again for giving you the run-around, but if you can put it back to using `control` instead of `control_class`, we can merge this PR and move on. Thanks in advance.

lcreid: BTW: I figured this out by adding a test in a local branch:
```
  test "control_class passed correctly" do
    expected = <<-HTML.strip_heredoc
      <div class="form-group">
        <label class="required" for="user_email">Email</label>
        <input class="my-class" id="user_email" name="user[email]" type="text" value="steve@example.com" />
      </div>
    HTML
    assert_equivalent_xml expected, @builder.text_field(:email, control_class: "my-class")
  end
```

lcreid:
Sorry. I didn't notice that this PR had been updated. @mattbrictson I'll take your comment as approval for this PR and merge it. Let me know if I'm overstepping my authority.

PR #448, opened 2018-03-08
Move validation errors for check boxes and radio buttons into last `div.form-check`.

### LINKED ISSUE #418: Errors on checkbox/radios collections should be inside the last element

lcreid:
Bootstrap 4 validation errors for groups of check boxes or radio buttons are supposed to be inside the `<div class="form-check">` of the last check box or radio button in the group. See: https://getbootstrap.com/docs/4.0/components/forms/#supported-elements.

This is the third variation on a solution to this issue

- PR #436 attempted to maintain backwards compatibility with blocks to `form_group` that rendered text other than from `bootstrap_form` helpers, but was never completely successful
- PR #422 did much of the work for this issue, but was abandoned because it didn't retain backwards compatabilty with `form_group` blocks rendering arbitrary text. This PR is heavily based on the approach used in PR #422, although it's actually a reimplementation

See https://github.com/bootstrap-ruby/bootstrap_form/pull/436#issuecomment-369350477 and the comments that follow it for why we decided to take the approach in this PR.

This PR also includes a UPGRADE-4.0.md document describing breaking changes and work-arounds that result from this PR.

Fixes #418.

### DIFF
--- a/test/bootstrap_form_group_test.rb
+++ b/test/bootstrap_form_group_test.rb
@@ -199,8 +199,8 @@ class BootstrapFormGroupTest < ActionView::TestCase
         <label class="col-form-label col-sm-2 required" for="user_email">Email</label>
         <div class="col-sm-10">
           <input class="form-control" id="user_email" name="user[email]" type="text" value="steve@example.com" />
+          <small class="form-text text-muted">This is required</small>
         </div>
-        <small class="form-text text-muted">This is required</small>
       </div>
     HTML
     assert_equivalent_xml expected, @horizontal_builder.text_field(:email, help: "This is required")


test/bootstrap_form_group_test.rb
```diff
@@ -359,23 +359,78 @@ class BootstrapFormGroupTest < ActionView::TestCase
...
+  #   end
+  #
+  #   expected = <<-HTML.strip_heredoc
+  #     <div class="form-group">
+  #       <p class="form-control-static">Bar</p>
+  #       <div class="invalid-feedback">can't be blank, is too short (minimum is 5 characters)</div>
+  #     </div>
+  #   HTML
+  #   assert_equivalent_xml expected, output
+  # end
```
mattbrictson: I'm going to remove this. It the comment served well for the review of this PR, but I would rather not keep comment-out code in the codebase. I'll move the reasoning for removing this into a commit comment instead.

mattbrictson:
@lcreid after rebasing this PR on the latest `master`, a test started failing. To get this PR passing I had to make the small change in the expected test output in commit 7c62989. Was this test failure legitimate? In which case a regression was introduced in this PR?

See test failure here: https://travis-ci.org/bootstrap-ruby/bootstrap_form/jobs/384966902#L913

lcreid:
@mattbrictson the correction you made to the test case is correct. I know the tests were all passing when I originally submitted the PR. My git-fu is not good enough for me to figure out how the test could have been passing when I originally submitted the pull request. (Actually, I have an idea how this happened, but I'll submit this comment and follow up later.)

The correction you made to the test case is in fact a change to the expected output that is a result of this PR. In other words, this PR changes the expected output, due to the way Bootstrap 4 works. 

As I said, the code is all good the way it is right now.

lcreid:
FWIW: PR #454 fixed the validation error for horizontal forms. I must have broken the test case again somehow, although I still can't explain why Travis didn't complain until @mattbrictson did the final merge. At any rate, it's all right now.

suketk:
Does this solve inline radio buttons? The formatting is wonky when you use error_message on the last radio.

Screen Shot 2022-08-25 at 7 09 33 PM

PR #450, opened 2018-03-09
Add text-danger to label when label_errors: true

lcreid:
This maintains the appearance of label errors consistent with v2.7 (Bootstrap 3).

With Bootstrap 4's way of marking up validation errors, the label text with a label error was not displaying in the error colour (default red). This PR adds `.text-danger` to a label when `label_errors: true` and there is a validation error on the field.

`invalid-feedback` was rejected as it changes the font appearance, and also doesn't appear to display, even though it's a sibling of the `input.is-invalid`. It looks like `.text-danger` is always the colour of validation errors, unless the programmer is doing some really interesting customization of Bootstrap colours.

No issue was entered for this as the fix was so quick. (My bad.)


mattbrictson:
Nice fix, thanks! 💯

PR #454, opened 2018-03-17
Moves help text in horiz. forms under the control

tekkla:
When using horizontal forms the help text is placed inside the created row but as sibling of the cols for label and control. This codechange moves the help text inside the col of the control so it gets aligned with/under the control.

Before:
from

After:
to


mattbrictson:
Looks great, thanks! 🙌 

I'll rebase and merge this once #455 is in.

tekkla:
Travis complains about this. 

> Fatal: Couldn't find remote ref refs/pull/454/merge

Do I have to change something or is it a problem on your side?

PR #455, opened 2018-03-20
Support for i18n :html subkeys in help text

jsaraiva:
This adds support for help text using :html subkey instead of the "_html" suffix to the attribute name.
Both are supported by i18n, actually: http://guides.rubyonrails.org/i18n.html#using-safe-html-translations .

This code just makes an additional check for an html subkey. If none is found, it continues in the regular fashion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
   [@duleorlovic](https://github.com/duleorlovic)
 * Your contribution here!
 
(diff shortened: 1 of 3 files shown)


jsaraiva:
Hi, @lcreid !
You're very welcome. :-)

It is possible that the change could benefit from some small refactoring, but I opted to keep it as simple as possible, for now.

lcreid:
@jsaraiva that was the right call. At the moment, we're just trying to make this work for Bootstrap 4, with minimal code changes. Thanks again!

mattbrictson:
@jsaraiva thanks, and congrats on your first contribution to the project! I will merge this in.

PR #456, opened 2018-03-25
Fix label `for` attribute when passing non-english characters using `collection_check_boxes`

ug0:
Fixed a problem that `FormBuilder#check_box_with_bootstrap` https://github.com/bootstrap-ruby/bootstrap_form/blob/master/lib/bootstrap_form/form_builder.rb#L140 will remove any non-English letters passed in as the `for` attribute for the label tag. Therefore any group of checkboxes contains non-English values won't work.


mattbrictson:
Nice fix, thanks!

Issue #460, opened 2018-04-28
inline option on date_select


lancecarlson:
The default output of the date_select fields when using horizontal layout on the form does not look good. The fields stack on top of each other. Perhaps there should be a way to specify inline: true on it so you can override this behavior?

lcreid:
Thanks for reporting this. We really appreciate your feedback.

Can you please confirm that [this Bootply](https://www.bootply.com/xIjouEAIwR) is what you're seeing?  I see what you mean, that it's not really what someone would expect on a horizontal form. 

Is [this Bootply](https://www.bootply.com/UVog5qkacr) more like what you'd like to see? It certainly seems a lot better to me this way.

lancecarlson:
@lcreid Yes! I think that's exactly what someone would be expecting. I haven't verified the actual look, but from an html perspective, this looks right. Thank you!

lcreid:
@lancecarlson are you interested in submitting a PR for this? We always appreciate contributions from the community. If not, no worries. I have some time this week and can probably put something together. Please just let me know what you like to do.

lancecarlson:
@lcreid please review! ^

lcreid:
This issue should probably include `time_select` and `datetime_select`, shouldn't it?

lcreid:
This was closed by PR #461.

PR #461, opened 2018-04-30
default form-inline class applied to parent content div on date select helpers

lancecarlson:
can also override this behavior with skip_inline: true

### DIFF
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -54,7 +54,9 @@ module BootstrapForm
       define_method(with_method_name) do |name, options = {}, html_options = {}|
         form_group_builder(name, options, html_options) do
           html_class = control_specific_class(method_name)
-          html_class = "#{html_class} form-inline" unless options[:skip_inline]
+          if @layout == :horizontal && !options[:skip_inline].present?
+            html_class = "#{html_class} form-inline"
+          end
           content_tag(:div, class: html_class) do
             input_with_error(name) do
               send(without_method_name, name, options, html_options)
--- a/lib/bootstrap_form/form_builder.rb
+++ b/lib/bootstrap_form/form_builder.rb
@@ -53,7 +53,9 @@ module BootstrapForm
 
       define_method(with_method_name) do |name, options = {}, html_options = {}|
         form_group_builder(name, options, html_options) do
-          content_tag(:div, class: control_specific_class(method_name)) do
+          html_class = control_specific_class(method_name)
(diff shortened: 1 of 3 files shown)


lancecarlson:
Fixes: https://github.com/bootstrap-ruby/bootstrap_form/issues/460

lib/bootstrap_form/form_builder.rb
```diff
@@ -52,7 +52,9 @@ def initialize(object_name, object, template, options)
       define_method(with_method_name) do |name, options = {}, html_options = {}|
         prevent_prepend_and_append!(options)
         form_group_builder(name, options, html_options) do
-          content_tag(:div, send(without_method_name, name, options, html_options), class: control_specific_class(method_name))
+          html_class = control_specific_class(method_name)
+          html_class = "#{html_class} form-inline" unless options[:skip_inline]
```
lcreid: I suspect that you will need to be more specific here about when to add `form-inline`. I believe you only want to add it when the `form_layout` is `:horizontal`.

lcreid: Is the `:skip_inline` option a new option you're proposing? If so, could you please update the CHANGELOG to note the new functionality, and probably add a description of the option in the `README`.

lancecarlson:
@lcreid I have no problem writing tests, but I'm having a hard time getting them to run. It keeps saying this:

ActiveRecord::StatementInvalid: Could not find table 'users'

lcreid:
Yes. I just figured that out myself. I'll fix the documentation.

lcreid:
Documentation fix in PR #462.

lancecarlson:
@lcreid I updated the changelog in this latest commit, but I'm not entirely sure where the skip_inline on the field helper should be documented. I only put it there so people could override if that wasn't the behavior they wanted.

lcreid:
Sorry for not following up on this sooner! This is looking good, but it occurs to me that we should do the same for `time_select` and `datetime_select`. Would you mind adding those to this PR similarly to the way you did `date_select`? Sorry I didn't mention this sooner.

lcreid:
@lancecarlson Please let me know if you're going to add this functionality for `time_select` and `datetime_select`. I can do it if you don't have time. Just let me know, please. Thanks in advance!

lcreid:
@lancecarlson thanks for the PR! My apologies for not noticing sooner that you had extended it to cover all the date and time selects. I resolved the conflicts.

@mattbrictson can you please take a quick look at this to make sure the PR doesn't add functionality that it shouldn't? The Bootstrap 4 version was stacking the fields for date and time, rather than putting them in the more intuitive side-by-side #460.

PR #462, opened 2018-05-04
Document database build before running tests.

lcreid:
The documentation wasn't clear enough about building the database before running tests, so it was easy for new contributors to miss the database creation. This PR hopefully clarifies what has to be done before running tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,9 +23,11 @@ following steps.
 
 For the project. Optionally, create a branch you want to work on
 
-### 3. Get it running locally
+### 3. Test it locally
 
-- Install the required dependencies with `bundle install`
+- The first time after you've cloned the repository:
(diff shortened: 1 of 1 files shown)


mattbrictson:
> Why was it removed?

Thanks for calling this out!

`ENV["RAILS_ENV"] = "test"` was removed from `test_helper.rb` in Rails 5.1.0 (see <http://railsdiff.org/5.0.7/5.1.0>), which is probably what our current test files are based on. That change has since been rolled back; Rails 5.2.0 has reinstated it. We should probably do the same.

I'll test to see if that fixes the issue being addressed by this PR.

mattbrictson:
Based on my testing, adding back the RAILS_ENV does indeed fix it. Thanks @GBH 🙌

mattbrictson:
Closing in favor of #464.

PR #463, opened 2018-05-10
Allow rendering of form submit buttons as HTML buttons, instead of input

jsaraiva:
As the title says, this PR allows devs to render submit buttons as HTML buttons, instead of using input tags.
The major advantage of this is that these submit buttons can then be styled, because they can have HTML content, as well as support for CSS pseudo-selectors.

All existing code should keep working normally. To use buttons instead of input tags, provide the `render_as_button` option with a truthy value:
`<%= f.primary t('.signin_button'), render_as_button: true %>`

Cheers!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ In addition to these necessary markup changes, the bootstrap_form API itself has
 * [#455](https://github.com/bootstrap-ruby/bootstrap_form/pull/455): Support for i18n `:html` subkeys in help text - [@jsaraiva](https://github.com/jsaraiva).
 * Adds support for `label_as_placeholder` option, which will set the label text as an input fields placeholder (and hiding the label for sr_only).
 * [#449](https://github.com/bootstrap-ruby/bootstrap_form/pull/449): Passing `.form-row` overrides default `.form-group.row` in horizontal layouts.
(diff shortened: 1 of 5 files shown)


jsaraiva:
Hi, Larry!

I've been thinking about your suggestion, and you're right 👍 . I had initially implemented it like that (in the submit helper) because I don't agree much with Rails' semantic treatment of the input helper. IMHO, a submit helper should be renderable as any element that HTML allows for submitting forms; currently, that is input[type="submit"] and button[type="submit"] (or button[type=""]) tags.

But, this would be a considerable departure from Rails' current helpers, and thus go against the philosophy of this great gem. :-)
Therefore, I have reverted the change to the submit helper and, following your advice, added a button helper (that just adds a CSS class if none is given, just like submit does).
I have however changed the primary helper so that, if it receives a :render_as_button option _or_ a content block, it will output a button instead an input tag.

Thanks for the tip!

jsaraiva:
Hi, Larry!

It's no problem 👍 . I should have done that to begin with. :-(

I also took the liberty of adding the `submit` helper to the list, and an explanation to the `primary`helper about the rendering as HTML button. The "save" button example used is similar to what I usually do with this new addition, but I typically use SCSS in conjunction with FontAwesome's source code to achieve the same effect. :-)

Let me know if you wish me to revert something, or if you need anything else.

Cheers!

lcreid:
@mattbrictson could you please take a look at this? I think @jsaraiva found a legit hole in our coverage of Rails helpers, and filled it nicely. Let me know if you want me to merge it, or merge it yourself if that's easier.

mattbrictson:
This looks good. I'll clean up the conflicts and merge. Thanks!

PR #465, opened 2018-06-01
error_summary: don't render surrounding div if there are errors

TomK32:
When there are no errors the div with for the error messages was still rendered, making it impossible to use css to style that surrounding div.

### DIFF
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -34,9 +34,11 @@ module BootstrapForm
       end
 
       def error_summary
-        content_tag :ul, class: 'rails-bootstrap-forms-error-summary' do
-          object.errors.full_messages.each do |error|
-            concat content_tag(:li, error)
+        if object.errors.full_messages.any?
+          content_tag :ul, class: 'rails-bootstrap-forms-error-summary' do
+            object.errors.full_messages.each do |error|
+              concat content_tag(:li, error)
+            end
           end
         end
       end
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -34,7 +34,7 @@ module BootstrapForm
       end
 
       def error_summary
-        if object.errors.full_messages.any?
+        if object.errors.any?
           content_tag :ul, class: 'rails-bootstrap-forms-error-summary' do
             object.errors.full_messages.each do |error|
               concat content_tag(:li, error)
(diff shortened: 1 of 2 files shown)


TomK32:
Right, I'm an RSpec guy ever since 2008 and someone else gotta tell me how to fix that test. No such test for `alert_message` _without_ errors btw.

lcreid:
Thanks for the PR. We always appreciate feedback and improvements to `bootstrap_form`. Thanks for catching that we have no test case for that method.

`error_summary` has worked this way for a long time, and I fear there would be code out there that depends on the current behaviour, e.g. the fact that it generates an empty `<ul></ul>`. 

I've been helping to support this gem for less than a year, so I can't speak for the original designers, but it looks like they intended for you to use the [`alert_message`](https://github.com/bootstrap-ruby/bootstrap_form#alert-messages) helper rather than the `error_summary` helper. It only outputs something if the object has errors .

If you really need to use `error_summary`, it seems to me that your use case could be handled fairly easily by writing this in a view:
```
<%= error_summary if @user.errors.full_messages.any? %>
```
If that was repetitive, you could write your own helper to do it. 

@TomK32 are either of those solutions acceptable to you, without changing the code in `bootstrap_form`?

By the way, your test case was good. The problem you were having is that `@user` is almost always used to generate error cases, so it's missing a required attribute. I added the attribute. Hopefully that makes the test pass.

@mattbrictson , can you please comment and correct me if I'm wrong on any of the above?

Thanks again for the contribution!

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -34,9 +34,11 @@ def alert_message(title, options = {})
       end
 
       def error_summary
-        content_tag :ul, class: 'rails-bootstrap-forms-error-summary' do
-          object.errors.full_messages.each do |error|
-            concat content_tag(:li, error)
+        if object.errors.full_messages.any?
```
mattbrictson: Could this be simplified to `object.errors.any?`, following the example of the Rails scaffold? https://github.com/rails/rails/blob/v5.2.0/railties/lib/rails/generators/erb/scaffold/templates/_form.html.erb.tt#L2

test/bootstrap_form_test.rb
```diff
@@ -572,6 +572,13 @@ class BootstrapFormTest < ActionView::TestCase
     assert_equivalent_xml expected, @builder.error_summary
   end
 
+  test "error_summary returns nothing if no errors" do
```
mattbrictson: 👍

lcreid:
@TomK32 Could you please make the changes requested by Matt above, and then we'll merge this PR. Thanks in advance.

PR #466, opened 2018-06-03
Generate empty but visible label

### LINKED ISSUE #457: Support nil value for label parameter
Looks like passing `nil` or an empty string to the label parameter does not remove the text, it shows the attribute name, I was wondering if it could be a good idea to support nil to remove that text, in my case I was trying to use the check_box helper method and trying to integrate it with a bootstrap switch plugin which only uses css, so it needs a label without content.

If this is something that can be merge, I'd love to create a PR for it.

Here an example

` = f.check_box :name`

```
<div class="form-check">
  <input name="name" type="hidden" value="0">
  <input class="form-check-input" type="checkbox" value="1" name="name" id="name">
  <label class="form-check-label" for="name">Active</label>
</div>
```

` = f.check_box :name, label: nil`

```
<div class="form-check">
  <input name="name" type="hidden" value="0">
  <input class="form-check-input" type="checkbox" value="1" name="name" id="name">
  <label class="form-check-label" for="name"></label>
</div>
```

lcreid:
Some third-party plugins require an empty but visible label on controls. Since `hide_label` makes the label invisible, another solution is required.

Since this seems like an edge case, rather than add another option, this PR proposes a documentation change to show how to generate an empty, visible label.

Test cases were added to confirm the functionality, but are not really required as this is a documentation-only change.

Fixes #457.


lcreid:
Never mind. I'll redo this PR.

Issue #347, opened 2017-09-13
Additional form group attributes don't work for date_select and time_select


katahreza:
Hi!
`f.date_select :my_date, wrapper: { class: 'additional-class' } `
or
`f.date_select :my_date, wrapper_class: 'additional-class'`

generate

`<div class="form-group">` without 'additional-class'

Wherein for text_field additional attributes are generated normally

pachacamac:
for others struggling .. this actually works:

`f.date_select :my_date, {start_year: Date.current.year, end_year: Date.current.year + 1}, {class: 'form-control'}`

lcreid:
Thanks for the suggestion, @pachacamac . I'm sure lots of people struggle with this, so your contribution is welcome. 

Could you please show us the generated HTML so that we can be sure what you mean by the "created child-elements"? Thanks in advance.

mattbrictson:
I'll consider this fixed by #473 and b41295e.

PR #473, opened 2018-07-03
Fix `wrapper_class` and `wrapper` options for helpers that have `html_options`.

lcreid:
`wrapper` and `wrapper_class` options were ignored for `date_select` and `time_select`. Further investigation indicated that the options were ignored for all helpers that take an explicit `html_options` argument.

The processing of `wrapper` and `wrapper_class` are moved to earlier in `form_group_builder`, and they're pulled explicitly from `options`, and not `html_options`. I believe this is correct, as the intent of `html_options` appears to be as a way to simply pass attributes to the underlying control.

I believe this fixes the original issue reported in #347, but may not address other comments on the issue.


mattbrictson:
Merged into master as b41295e

PR #483, opened 2018-08-22
The `button`, `submit`, and `primary` helpers can now receive an addi…

jsaraiva:
The `button`, `submit`, and `primary` helpers can now receive an additional option, `extra_class`. This option allows us to specify additional CSS classes to be added to the corresponding button/input, _while_ maintaining the original default ones. E.g., a primary button with an `extra_class` 'test-button' will have its final CSS classes declaration as 'btn btn-primary test-button'.

This is particularly useful for adding extra details to buttons without forcing you to repeat the
Bootstrap classes. You can just write `extra_class: 'test-button'` instead of `class: 'btn btn-primary test-button'`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### New features
 
 * Your contribution here!
+* The `button`, `submit`, and `primary` helpers can now receive an additional option, `extra_class`. This option allows us to specify additional CSS classes to be added to the corresponding button/input, _while_ maintaining the original default ones. E.g., a primary button with an `extra_class` 'test-button' will have its final CSS classes declaration as 'btn btn-primary test-button'.
 
 ### Bugfixes
 
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ this defining these selects as `inline-block` and a width of `auto`.
 
 ### Submit Buttons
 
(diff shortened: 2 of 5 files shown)


jsaraiva:
Hi, Larry!

Thank you very much. :-)

I only suggested this because I have had a **lot** of cases in which I used
```erb
f.button 'some text', class: 'btn btn-primary my-special-css-class'
```
when it would have been more readable to just do
```erb
f.primary 'some text', class: 'my-special-css-class'
```
A typical example of _my-special-css-class_ is _save_, for the main "Save changes" button in a form.
Same goes for the `button` helper, with 'btn btn-secondary'.

However, the `class` attribute is reserved, and for good reason.
Of course, it could be argued that, in `primary`'s case, it should always add the _btn btn-primary_ classes (because otherwise we would just use the `button` helper), but backward-compatibility is very important (I know; I confess I was caught off-guard by the change to the wrapper class in the `select` helper), and this was the only way I could figure out to achieve my goals -- simplify my code -- without breaking anything.

BTW, I first tried just using
```
options.reverse_merge! class: "btn btn-primary #{options.delete(:extra_class)}"
```
but it broke a lot of tests because of the extra space char in the class declaration. If you wish to remove the private helper method (and change the other tests to reflect that space), that should work too.

Cheers!

PR #488, opened 2018-10-16
add required option on form_group_builder

ThomasSevestre:
Add required option on form_group_builder :

            = f.text_field :firstname, required: true

This option forces required class to be added, even if the there is no presence validation on the model.
This is usefull when the object model is not an active record model.

I will add tests and doc if you are interested

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,12 @@ In cases where this behavior is undesirable, use the `skip_required` option:
 <%= f.password_field :password, label: "New Password", skip_required: true %>
 ```
 
+In cases where the label `required` class should be added but the associated model has no presence validation, use the `required` option:
+
+```erb
+<%= f.password_field :password, label: "New Password", required: true %>
+```
+
 ### Input Elements / Controls
 
 To specify the class of the generated input tag, use the `control_class` option:
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ validator with the associated model attribute. Presently this is one of:
 ActiveRecord::Validations::PresenceValidator or
 ActiveModel::Validations::PresenceValidator.
 
-In cases where this behavior is undesirable, use the `required` to force the class to be present or absent :
+In cases where this behavior is undesirable, use the `required` option to force the class to be present or absent:
 
 ```erb
 <%= f.password_field :login, label: "New Username", required: true %>
--- a/README.md
+++ b/README.md
@@ -213,16 +213,11 @@ validator with the associated model attribute. Presently this is one of:
 ActiveRecord::Validations::PresenceValidator or
 ActiveModel::Validations::PresenceValidator.
 
-In cases where this behavior is undesirable, use the `skip_required` option:
+In cases where this behavior is undesirable, use the `required` to force the class to be present or absent :
 
 ```erb
-<%= f.password_field :password, label: "New Password", skip_required: true %>
-```
-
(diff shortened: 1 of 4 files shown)


lcreid:
I'd like to submit a PR to apply some Rubocop style rules to the entire code base of `bootstrap_form`. Do you think you'll be doing the tests and documentation for this PR in the next few days. If so, I can wait. If not, I'd prefer to merge the PR for Rubocop ASAP, so that we can all start working with code that looks a little more consistent.

README.md
```diff
@@ -219,6 +219,12 @@ In cases where this behavior is undesirable, use the `skip_required` option:
 <%= f.password_field :password, label: "New Password", skip_required: true %>
 ```
 
+In cases where the label `required` class should be added but the associated model has no presence validation, use the `required` option:
```
lcreid: Please add a line that says the `required: false` does nothing, or, change the code so that `required: false` is equivalent to `skip_required: true`. If you change the code, don't forget to add a test case for `required: false`. If you just change the documentation, I don't think you need to add a test case.

I don't have an opinion about which is the right solution. Since you have the use case, I'm happy to leave it to your judgement.

ThomasSevestre: Good catch !

To simplify the API I would do the following : 
- deprecate skip_required
  * add a warning (like rails is doing)
  * raise an error if skip_required and the new required options are used at the same time
- allow 3 values on the new required option:
  * `:infered` : default behavior (required class is added if validation is present)
  * `false` : required class NOT added, whatever the model validation
  * `true` : required class added, whatever the model validation

Or I'll add a line in the readme to say that `required: false` does nothing.

lcreid: I love the enthusiasm. If you want to modify your PR for what you listed under "To simplify the API...", it sounds good to me. 

The only suggestion I have is that we make it a warning rather than an error if both `skip_required` and `required` are used, and use the `skip_required` option. This would prevent breaking sites where someone in the past:

1. Used `required` without looking in the docs.
2. When it didn't work, saw that the option was `skip_required`.
3. Added `skip_required` but forgot to remove the `required` option.

I know that's a bit extreme, but I think this case warrants a warning rather than raising an exception.

One small detail: "inferred" has two "r". (And no need to apologize. English spelling sucks. I had to look it up myself to be sure.) Or you could just use `:default` instead of `inferred`. I think some other options use `:default`.

lcreid:
Let me look at the Rubocop rules to try to fix the Travis build. I don't think there's anything wrong with your code. We'll probably merge this PR as long as it's only the Rubocop check that's failing.

lib/bootstrap_form/form_builder.rb
```diff
@@ -463,8 +464,8 @@ def generate_label(id, name, options, custom_label_col, group_layout)
         classes << "mr-sm-2"
       end
 
-      unless options.delete(:skip_required)
-        classes << "required" if required_attribute?(object, name)
+      if options.delete(:required) || (!options.delete(:skip_required) && required_attribute?(object, name))
```
lcreid: I take back my earlier comment. Could you please change these lines to the modifier form (`if` at the end of the line). That may fix all the RuboCop offences (even if I wish we didn't have such long lines).

lcreid:
Well done! Apologies for my sloppy attempts to fix some of the RuboCop offences. I thought I'd fix them myself rather than nagging you. Your work was excellent. Thanks for all your effort on this PR.

PR #498, opened 2019-01-03
Add example for changing default value for label_col and control_col

duleorlovic:
I added example for changing default `label_col` and `control_col` since default `col-sm-2` and `col-sm-10` does not look good when I show the form in thin modal.
Please do not merge if this is against some best practices (maybe there is a better way of changing default values, instead reopening BoostrapForm::FormBuilder class).

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,6 +540,22 @@ The `label_col` and `control_col` css classes can also be changed per control:
 <% end %>
 ```
 
+or default value can be changed in initializer:
+
+```erb
+# config/initializers/bootstrap_form.rb
+module BootstrapForm
+  class FormBuilder
+    def default_label_col
+      'col-sm-4'
+    end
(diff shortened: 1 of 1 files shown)


lcreid:
Thanks for the PR improving the documentation! It's a good idea to add this example. 

The original Travis build failed, but obviously the issue was unrelated to your documentation change. I believe I've fixed the Travis problem -- it's running right now. As soon as it completes I'll merge this.

Do you have think there are other options that might benefit from having a documented way to change the default?

duleorlovic:
I did not need to change any other option... On some projects I added new elements like:

* text field with select options (for example: `Download limit [ number input ] [ select KB/MB/GB ]`) so maybe that is useful to share, but it is not for README... I can not find Wiki on this project.

PR #513, opened 2019-01-31
Add support for `rich_text_area`.

### LINKED ISSUE #508: Add support for Action Text (Rails 6)

lcreid:
Rails 6 introduces the Trix editor for text areas with formatting and images.

The implementation in this PR works around some challenges in ensuring that the Rails `rich_text_area` helper was available when `bootstrap_form` needed it. I eventually had to put an explicit require of part of `ActionText` if running Rails 6. I don't like that, but I couldn't find another way to do it.

Because it was so hard to get the test to run, I made a separate test file for rich text areas. This let me control what was being included and when, so that I could make sure the testing was being done like the actual Rails testing. This tests should run fine now. It was just challenging to get all the pieces in place to make them work.

Improvements to this implementation are welcome.

Fixes #508.


Gemfile
```diff
@@ -3,12 +3,15 @@ source "http://rubygems.org"
 gemspec
 
 # Uncomment and change rails version for testing purposes
-# gem "rails", "~> 5.2.0.beta2"
+# gem "rails", "~> 5.0.0"
```
markburns: Might be a good use case for the [appraisals gem](https://github.com/thoughtbot/appraisal)

lcreid: We used to use the appraisals gem. Not too long before I became the maintainer, we stopped. I don't remember the reasoning for dropping appraisal. 

Another option that might help us is the https://github.com/clio/ten_years_rails gem, which helps you run the current and next future Rails version.

PR #514, opened 2019-01-31
Add support for BS 4.2 switches

simmerz:
Add :switch value to :custom option for 4.2 switches

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -18,7 +18,7 @@ Layout/IndentHeredoc:
 
 # Offense count: 9
 Metrics/AbcSize:
-  Max: 67
+  Max: 69
 
 # Offense count: 2
 # Configuration parameters: CountComments.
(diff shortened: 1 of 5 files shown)


simmerz:
@lcreid I've updated the documentation to show how to use the new :switch syntax

simmerz:
@lcreid this PR is now rebased against current master, and I've also refactored the `check_box.rb` file.

I had a question about double quotes - the `.rubocop.yml` file enforces double quotes, but the default is single quote unless interpolation is required. Any reason for this particular decision?

lcreid:
> @lcreid this PR is now rebased against current master, and I've also refactored the `check_box.rb` file.
> 
> I had a question about double quotes - the `.rubocop.yml` file enforces double quotes, but the default is single quote unless interpolation is required. Any reason for this particular decision?

I think the simple answer to your question is that both the previous maintainer and myself prefer double quotes. In my case, I like to use contractions when I write, so defaulting to double quotes for strings means I don't have to change the quotes every time I add or take away a contraction from a string. That said, I recognize that there are down sides to both options, so I hope it's not too much of an inconvenience for you.

lcreid:
Thanks so much for the PR! I really appreciate it that others are keeping up with changes to Bootstrap.

simmerz:
@lcreid not at all. And regarding double quotes - no inconvenience at all. Always better to ask and to learn than to just accept the status quo though.

Issue #517, opened 2019-01-31
Move inputs into separate files to make code more maintainable


simmerz:
This *should* also, in theory, reduce the scope of `rubocop_todo.yml`

This should maintain existing compatibility (ie no breaking changes) with the existing latest cut gem.

simmerz:
@lcreid closing this now. I should probably have done it in my commit message :|

PR #518, opened 2019-02-08
Separation of inputs into separate more maintainable files

simmerz:
@lcreid this is how I propose to tackle #517 .

I'd like to get this done across all the fields and then figure out if we can DRY any of it up, but perhaps that's a second exercise.

What are your thoughts?


lib/bootstrap_form/inputs/text_field.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
```
lcreid: Do you expect there will be one file for each of the FIELD_HELPERS? Did you consider using one file and one method to define the "typical" helpers listed in FIELD_HELPER? The current implementation makes it clear that the intent is that all those helpers are basically the same, but it's harder to see because of the Ruby meta-programming.

simmerz: So my short term intent is to get it working. My shorter term intent is to either subclass or use metaprogramming. I'm leaning towards subclassing though. It feels cleaner and easier to read. we can use class-level attribute to determine the field name `self.field_name = :text_field` for instance.

lcreid: I'm not sure where what you mean. If you don't mind the risk that we might decide not to go in that direction, why don't you continue working on this PR, so I get a better idea of where you're going. 

Did you have any ideas about what to do with `form_group` and `form_group_builder`, and the methods below them?

simmerz:
@lcreid This has now been updated (it needs rebasing and the new `rich_text_field` adding, but before I do that I wanted to get approval on this approach).

I've taken the liberty of removing the method_alias_chain "clone" file as it can be done more simply as I've demonstrated in this.

I've deliberately not refactored the various field inputs into smaller ones as I think that is a separate job.

lib/bootstrap_form/inputs/text_field.rb
simmerz: @lcreid to answer your questions about form_group/form_group_builder, I think this is a problem best tackled separately.

lcreid:
I like what I see. There's a lot of work here, and you obviously have a good grasp of loading, requiring, etc. I've wondered myself whether there were better ways to do things with the new Ruby and Rails that we're supporting nowadays. I also like how you cleaned up the redundancy around the methods that are similar. 

Still, I'd like to take a longer look at it (sorry for the delay already, but paid work has to come first, I'm afraid). 

In the meantime, I've realized that we didn't have any tests to test for the "_without_bootstrap" methods. Since you've changed the implementation of that, it would be worthwhile to add at least one test of a "_without_bootstrap" method, if for no other reason, than to record that it's a requirement. E.g. 
```
test "text fields are wrapped correctly" do
  expected = <<-HTML.strip_heredoc
    <input id="user_email" name="user[email]" type="text" value="steve@example.com" />
  HTML
  assert_equivalent_xml expected, @builder.text_field_without_bootstrap(:email)
end
```
That may not be exactly right, but should give you the idea of what I'm suggesting.

Thanks again. Great work!

lib/bootstrap_form/form_builder.rb
```diff
@@ -23,241 +46,16 @@ def initialize(object_name, object, template, options)
...
                          @label_errors != true
                        else
                          options[:inline_errors] != false
                        end
       @acts_like_form_tag = options[:acts_like_form_tag]
-
       super
     end
 
-    FIELD_HELPERS.each do |method_name|
```
lcreid: I'm curious why you chose to remove the enumeration here and on line 51? I like what you did in that it's consistent, but I also like the fact that I can quickly tell from the code which methods behave the same with respect to their options.

lib/bootstrap_form/inputs/check_box.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
```
lcreid: Can you please just confirm that you didn't change any of the code in this file, `radio_button.rb`, the code that's left in `form_builder.rb`, and anywhere else where you were copying and pasting, other than the bare minimum you needed to change to move the code?

lib/bootstrap_form/form_builder.rb
simmerz: We could have left it in, and I did consider it. My decision stemmed from being able to clearly see what field inputs there are. To the user, it doesn't matter what the implementation is (mostly). Just that it works, and by seeing what field types there are clearly, it helps with understanding how to use the gem.

lib/bootstrap_form/inputs/check_box.rb
simmerz: Confirmed. The sole change outside of this was in removing the method_chain_alias code and implementing the cleaner equivalent in `base.rb` as `bootstrap_alias`

simmerz:
> In the meantime, I've realized that we didn't have any tests to test for the "_without_bootstrap" methods. Since you've changed the implementation of that, it would be worthwhile to add at least one test of a "_without_bootstrap" method, if for no other reason, than to record that it's a requirement. E.g.

Added a test just like this

lib/bootstrap_form/inputs/check_box.rb
lcreid: Thanks!

lib/bootstrap_form/form_builder.rb
lcreid: That makes sense. I totally agree that the implementation shouldn't matter to the user (programmer using bootstrap_form). I like what you've done.

PR #522, opened 2019-02-19
Clean up rubocop offenses

simmerz:
This PR cleans up various rubocop offenses

I've also regenerated the `.rubocop-todo.yml` so it's up to date and doesn't contain things that have been fixed by cleaning up code elsewhere.


simmerz:
@lcreid I've done a load of work refactoring here and am now left with 2 files that I think should be in `test/` - `bootstrap_deferred_builder_test.rb` and `bootstrap_error_rendering_test.rb`

Is there any reason those are in the root of the project?

lcreid:
Ouch! Those were test files that I was using locally at some point. I obviously should have moved them out of the tree completely, because somewhere along the line I committed them by accident. Sorry. Feel free to remove them. Or let me deal with it later.

simmerz:
Ah ok. I did wonder seeing as I couldn't see how they could possibly pass! I'll remove them in a bit and then we're done with the to-do file.

simmerz:
@lcreid all done

lib/bootstrap_form/inputs/file_field.rb
```diff
@@ -12,21 +12,24 @@ def file_field_with_bootstrap(name, options={})
...
               input_with_error(name) do
-                placeholder = options.delete(:placeholder) || "Choose file"
-                placeholder_opts = { class: "custom-file-label" }
-                placeholder_opts[:for] = options[:id] if options[:id].present?
-
-                input = file_field_without_bootstrap(name, options)
-                placeholder_label = label(name, placeholder, placeholder_opts)
-                concat(input)
-                concat(placeholder_label)
+                file_filed_input(name, options)
```
lcreid: Typo. I think you want "field", not "filed".

lib/bootstrap_form/inputs/file_field.rb
```diff
@@ -12,21 +12,24 @@ def file_field_with_bootstrap(name, options={})
...
             end
           end
         end
 
         bootstrap_alias :file_field
       end
+
+      private
+
+      def file_filed_input(name, options)
```
lcreid: Typo. I think you want "field", not "filed".

lib/bootstrap_form/inputs/check_box.rb
```diff
@@ -65,16 +68,22 @@ def check_box_label_class(options)
       def check_box_wrapper_class(options)
         classes = []
         if options[:custom]
-          classes << "custom-control"
-          classes << (options[:custom] == :switch ? "custom-switch" : "custom-checkbox")
-          classes << "custom-control-inline" if layout_inline?(options[:inline])
+          classes << custom_check_box_wrapper_class(options)
         else
           classes << "form-check"
```
lcreid: For radio buttons you defined a `standard_radio_button_wrapper_class`. Did you want to define a `standard_check_box_wrapper_class`? I don't care which way you do it, but I thought I'd mention it in case you meant to and forgot.

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -63,7 +62,7 @@ def static_control(*args)
 
         static_options = options.merge(
           readonly: true,
-          control_class: [options[:control_class], static_class].compact.join(" ")
+          control_class: [options[:control_class], static_class].compact
```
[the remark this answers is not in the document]
simmerz: `class` attributes in Rails accept arrays and do the join internally, so it makes it easier to just present an array and push into it as needed.

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -80,14 +79,13 @@ def custom_control(*args, &block)
 
       def prepend_and_append_input(name, options, &block)
         options = options.extract!(:prepend, :append, :input_group_class)
-        input_group_class = ["input-group", options[:input_group_class]].compact.join(" ")
 
-        input = capture(&block) || "".html_safe
+        input = ActiveSupport::SafeBuffer.new(capture(&block) || "")
```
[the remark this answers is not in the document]
simmerz: Oops! Yep, totally agree with your approach here.

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -108,13 +106,23 @@ def static_class
 
       private
 
+      def append_input(options)
+        html = content_tag(:div, input_group_content(options[:append]), class: "input-group-append") if options[:append]
+        ActiveSupport::SafeBuffer.new(html || "")
```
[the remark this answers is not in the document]
simmerz: True enough. Also, no need for the `""` as `.new` does the same thing

lib/bootstrap_form/inputs/check_box.rb
simmerz: I was refactoring specifically to meet rubocop requirements, rather than necessarily to make them behave in the same way, hence I did it like this.

simmerz:
@lcreid Updates pushed. Thanks for the review!

lib/bootstrap_form/inputs/check_box.rb
lcreid: OK. Thanks.

lib/bootstrap_form/helpers/bootstrap.rb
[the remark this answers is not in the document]
lcreid: I had occasion to try it in another app last night, and felt a little sheepish that I'd never known that before. :-) Thanks!

lcreid:
Once again, a big thanks for all the work you've done on this gem!

PR #524, opened 2019-02-26
Only concat help_text if it exists

simmerz:
@lcreid I found a bug in my own code. This little fix allows help_text to be nil, and still work :|

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -29,7 +29,7 @@ end
 # Have you updated CHANGELOG.md?
 # ------------------------------------------------------------------------------
 if !has_changelog_changes && has_lib_changes
-  markdown <<-MARKDOWN.strip_heredoc
+  markdown <<~MARKDOWN
     Here's an example of a CHANGELOG.md entry (place it immediately under the `* Your contribution here!` line):
 
     ```markdown
--- a/Dangerfile
+++ b/Dangerfile
@@ -29,12 +29,12 @@ end
 # Have you updated CHANGELOG.md?
 # ------------------------------------------------------------------------------
 if !has_changelog_changes && has_lib_changes
-  markdown <<~MARKDOWN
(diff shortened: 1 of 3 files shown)


Dangerfile
```diff
@@ -29,7 +29,7 @@ end
 # Have you updated CHANGELOG.md?
 # ------------------------------------------------------------------------------
 if !has_changelog_changes && has_lib_changes
-  markdown <<-MARKDOWN.strip_heredoc
+  markdown <<~MARKDOWN
```
lcreid: I believe this change breaks the Travis build because we still have to support Ruby 2.2. Could you please put this file back the way it was, if that's indeed the problem? Thanks in advance.

Sorry for the double review. I hit approve before I saw that Travis was failing.

simmerz: No worries. I noticed that the original with strip_heredoc was failing too on one of the builds. Perhaps just drop the strip_heredoc command?

lcreid: I haven't run into a `strip_heredoc` failing before. Do you recall what version or what test failed with the `<-MARKDOWN.strip_heredoc` version? In the meantime, could you try reverting this one change from the PR so it passes Travis and I can merge it? With Rails 6 coming out soon, we'll be reviewing all the versions of Ruby and Rails that we support, and we can clean up some of the old Ruby syntax at that time, if it causes a problem.

In my testing, the Dangerfile passes RubyCop with the original syntax, because RuboCop is configured to use Ruby 2.2 syntax. The Dangerfile itself runs fine with the new syntax, because `.travis.yml` says run Danger with Ruby 2.5.

lcreid: @simmerz would you mind fixing this PR so it passes Travis? I'd like to do a minor release of the gem, but we should get this fix in before we release it. Thanks in advance.

lcreid:
Thanks!

PR #530, opened 2019-05-13
Dependency updates and lint

antpaw:
I've installed the project in on a "clean" system with the newest versions of ruby, sqlite ect. I could not run the tests, so this upgrade helped me to run it. Maybe you are interested.


lcreid:
Thanks for the PR. I like the way you cleaned up the check_box code. Sorry for the delay merging it.

PR #532, opened 2019-05-18
Update collection_check_boxes.rb

### LINKED ISSUE #531: collection_check_boxes with custom: true doesn't allow select
the issue is present because the collection_check_boxes method doesn't generate id's for the input elements nor a for attribute for the label which breaks the check box.

My suggestion is as follows: if a custom id is not present generate an id as follows "#{input_name}_#{value}" to ensure it being unique

I looked into the source code and couldn't find the origin of the problem. if someone can guide me i'll be happy to submit a PR

Dreamersoul:
set ids for collection_check_boxes if not supplied. this allows custom checkboxes to work and it is an html5 best practice for usability purposes. this fixes #531


Dreamersoul:
i have checked the tests and it indeed sets id's but i don't know why in my project it isn't

PR #537, opened 2019-07-02
Add some missing commas in the examples for static controls

yi-jiayu:


lcreid:
Thanks so much for the PR! Good catch!

Issue #540, opened 2019-07-25
Rich text area not auto-expanding vertically


markburns:
A picture is worth a thousand words.

------
![image](https://user-images.githubusercontent.com/104158/61865933-1b887580-aecc-11e9-99c8-03b4f0c4abeb.png)
------

I haven't dug into the CSS yet to understand what's happening, but I'm not sure this is working out of the box as you might expect it to.

markburns:
seems to be some clash with the `.form-control` class.

Workaround:

```slim
= form.label :summary
= form.hidden_field :summary
trix-editor id="foo" input="foo_trix_input_actors_my_profile_form" role="textbox"
  == form.object.summary
```

lcreid:
Thanks for the report. I will try to take a look at this in the near future, but I'm pretty busy at work these days. If you want to submit a pull request, I'd be happy to look at it. If you decide to work on a PR, assign the bug to yourself (look for the link in the upper-right area of this page.

PR #544, opened 2019-09-02
Rails 6

lcreid:
Update `Gemfile` to Rails 6.0.0. 

Update `.travis.yml` for the new support matrix, although leaving some earlier Rails versions in the matrix for now, since the Rails documentation is unclear about which old versions are being supported.

Add on-going testing against Rails edge, but don't let failures fail the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ gemfile:
   - test/gemfiles/5.0.gemfile
   - test/gemfiles/5.1.gemfile
   - test/gemfiles/5.2.gemfile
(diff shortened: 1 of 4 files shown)


kylefox:
Will there be a new gem release with Rails 6 compatibility?

By the way, thanks for your work on this gem @lcreid 👍

lcreid:
Version 4.3.0 has been deemed the official Rails 6.0.0 version. There is at least one reported issue with the rich text area support still (#540).

PR #548, opened 2019-10-21
Improve installation instructions

wynksaiddestroy:
The official bootstrap installation guide encourages using the SCSS pre-processor. So I added the required steps to include the stylesheet, if you use SCSS instead of vanilla CSS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,12 +47,16 @@ Then:
 
 `bundle install`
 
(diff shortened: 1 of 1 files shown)


wynksaiddestroy:
Thanks for the quick merge and the good work you're doing ❤

PR #549, opened 2019-10-26
Add support for Ransack::Search object to the first argument of bootrap_form_for

### LINKED ISSUE #546: Undefined method `name' for "Search":Ransack::Name
When using a ransack object, there's an error trying to get the model name's name:

https://github.com/bootstrap-ruby/bootstrap_form/blob/aa2cb2933cd655bb0ebf1e9691b28760f8a77054/lib/bootstrap_form/components/hints.rb#L23-L27

The workaround is setting the help text manually:
```
= f.text_field :nombre_cont, help: ""
```

harada4atsushi:
Fixes #546 

Add Support for Ransack::Search object to the first argument of bootrap_form_for. 

## Usage example:

Controller

```
def index
  @q = Invoice.ransack(params[:q])
  @invoices = @q.result(distinct: true)
end
```

View

```
= bootstrap_form_for @q, url: invoices_path, method: :get do |f|
  = f.search_field :file_name_cont
  = f.submit
```

### DIFF
--- a/lib/bootstrap_form/components/hints.rb
+++ b/lib/bootstrap_form/components/hints.rb
@@ -39,7 +39,13 @@ module BootstrapForm
       end
 
       def object_class
-        object.class.to_s == "Ransack::Search" ? object.klass : object.class
+        if object.class.is_a?(ActiveModel::Naming)
+          object.class
+        elsif object.respond_to?(:klass) && object.klass.is_a?(ActiveModel::Naming)
+          object.klass
+        else
+          object.class
+        end
       end
 
       def scoped_help_text(name, partial_scope)
--- a/lib/bootstrap_form/components/hints.rb
+++ b/lib/bootstrap_form/components/hints.rb
@@ -20,10 +20,10 @@ module BootstrapForm
       def get_help_text_by_i18n_key(name)
         return unless object
 
-        partial_scope = if object.class.respond_to?(:model_name)
-                          object.class.model_name.name
+        partial_scope = if object_class.respond_to?(:model_name)
+                          object_class.model_name.name
                         else
-                          object.class.name
+                          object_class.name
                         end
 
         # First check for a subkey :html, as it is also accepted by i18n, and the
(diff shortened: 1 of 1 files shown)


lib/bootstrap_form/components/hints.rb
```diff
@@ -38,6 +38,10 @@ def get_help_text_by_i18n_key(name)
         help_text
       end
 
+      def object_class
+        object.class.to_s == "Ransack::Search" ? object.klass : object.class
```
jits: Would it make sense to do something like:

```ruby
object.klass || object.class
```

?

This way the _contract_ is: if `object.klass` is defined, it takes precedence over the actual `class`.

harada4atsushi:
@lcreid 
Thank you for your reply.
I also agree that opinion.

Therefore, Ransack objects are not referenced directly, 
compare `object.klass` with string 'Ransack::Search', exceptions will not occur even if there is a change on the Ransack side.

should I avoid using external class names as strings?

I'll consider another idea.

lib/bootstrap_form/components/hints.rb
harada4atsushi: Thank you for the advice.
I think it is a very good idea. I also considered of the same idea first.

However, the implementation of `object.klass` is free, I felt that there is a possibility that an unintended value would come in and break the code on the library user side. So only in the case of Ransack to limit the impact.

harada4atsushi:
I Modified not to use the string `Ransack::Search`.
What about this code?(0797262) I will add tests later.

```
      def object_class
        if object.class.is_a?(ActiveModel::Naming)
          object.class
        elsif object.respond_to?(:klass) && object.klass.is_a?(ActiveModel::Naming)
          object.klass
        else
          object.class
        end
      end
```

donv:
@harada4atsushi Do you still want this change?  If so, could you add a test for it?

donv:
Hi @harada4atsushi !

Thanks for the PR!  If you would like it merged, please add a test that demonstrates the usage.  If not, I will close the PR.

PR #550, opened 2019-10-27
Add `default_layout` so we can use it for all forms

duleorlovic:
Move `process_options` from helper since we can not override helper.
When someone wants to extend `BootstrapForm::FormBuilder` and use it in
`form_with .... builder: MyFormBuilder`, before this change he could not
use in this way since determination whether it is :inline :horizontal
was done in helper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,6 +621,10 @@ module BootstrapForm
     def default_control_col
       'col-sm-8'
     end
+    def default_layout
+      # :default, :horizontal or :inline
+      :horizontal
+    end
   end
 end
 ```
--- a/lib/bootstrap_form/action_view_extensions/form_helper.rb
+++ b/lib/bootstrap_form/action_view_extensions/form_helper.rb
@@ -20,8 +20,6 @@ module BootstrapForm
       def bootstrap_form_for(record, options={}, &block)
         options.reverse_merge!(builder: BootstrapForm::FormBuilder)
 
-        options = process_options(options)
-
         with_bootstrap_form_field_error_proc do
           form_for(record, options, &block)
         end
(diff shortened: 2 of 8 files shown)


duleorlovic:
The change in test is only about adding class `new_user` which rails adds automatically for `form_for` syntax if  `html: { class: }` were not given in [apply_form_for_options](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L442) (please note [reverse_merge](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L466) and actual source for [new_name](https://github.com/rails/rails/blob/master/actionview/lib/action_view/record_identifier.rb#L76) class).

So I need to change only those tests that were using `layout: :inline` option for form_for since now we add `form-inline` in builder (not as `html: { class: 'form-inline' }` option for form_for).

If you like, I replaced those tests with newer `form_with` syntax and than we do not have those additional classes. I will revert last commit if `form_with` syntax is not needed.

I think it is safe to merge since this pull request will only add `new_user` class for `layout: :inline` forms (so form will have `new_user form-inlne` instead of `form-inline` class) and only for old `form_for` syntax.

duleorlovic:
I reverted `Use form_with syntax` since is it not available on Rails < 5.1...

lcreid:
I am comfortable that the change in test cases introduced by the PR is in fact the way they should be. If we break any existing users' code (which I think will be in relatively few cases), I'm comfortable defending it as a fix to code that was broken. 

@duleorlovic I'll merge this now, but please confirm that we don't need to change any documentation (I don't think so, but please forgive me for double-checking.)

duleorlovic:
I think example in README is enough https://github.com/bootstrap-ruby/bootstrap_form/pull/550/files#diff-04c6e90faac2675aa89e2176d2eec7d8R624

PR #555, opened 2020-01-23
FormGroup: don't concat help_text if nil

dunn:
Followup to #524, on the other branch of the conditional.


dunn:
not sure if this is actually necessary, after all

PR #559, opened 2020-03-06
Cache bundler to speed up travis build

donv:
Travis uses about 40 seconds to install gems.  The goal is to reduce this as much as possible.


donv:
Reduced build time from 26 to 11  minutes in travis!

PR #562, opened 2020-04-14
Allow to configure default form attributes

### LINKED ISSUE #561: Implement Configuration Options
One of the good features of `bootstrap_form` is that it doesn't require a complicated configuration process before you can use it.

Also, configuration options can greatly increase the development challenge, as you may need test cases for each variation of a configuration option.

On the other hand, there are a lot of good ideas for the future of `bootstrap_form` that require breaking changes. It would be relatively easy to introduce new features if they could be controlled by an configuration option, so that people who needed the legacy behaviour could retain it easily.

This issue is asking for proposed PRs for a configuration mechanism for `bootstrap_form`. The mechanism is likely a file in `config/initializers`, but we're open to other ideas if they have significant benefits.

If you choose to work on this ticket, please write up a brief description of what you plan to do, and assign the issue to yourself (with the "Asignees" like above and to the right).

sharshenov:
This PR adds ability to configure `bootstrap_form` default form options

```ruby
# config/initializers/bootstrap_form.rb
BootstrapForm.configure do |c|
  c.default_form_attributes = { something: "custom" }
end
```

Current default value is `{role: "form"}` which is not semantically correct as described in #560. In order to resolve original issue, `default_form_attributes` should be set to `{}`

```ruby
BootstrapForm.configure do |c|
  c.default_form_attributes = {}
end
```

Closes #561
Closes #560

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,12 +153,20 @@ in `form_with`.
 
 ## Configuration
 
-`bootstrap_form` has some default options that can be overriden:
+`bootstrap_form` can be used out-of-the-box without any configuration. However, `bootstrap_form` does have an optional configuration file at `config/initializers/bootstrap_form.rb` for setting options that affect all generated forms in an application.
 
+The current configuration options are:
+
+| Option | Default value | Description |
+|---------------------------|------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `default_form_attributes` | `{role: "form"}` | version [2.2.0](https://github.com/bootstrap-ruby/bootstrap_form/blob/master/CHANGELOG.md#220-2014-09-16) added a role="form" attribute to all forms.   The W3C validator will raise a **warning** on forms with a role="form" attribute.   Set this option to `{}` to make forms be W3C compliant. |
+
+
+Example:
 ```ruby
 # config/initializers/bootstrap_form.rb
 BootstrapForm.configure do |c|
-  # c.default_form_attributes = { role: "form" }
+  c.default_form_attributes = {} # to make forms W3C compliant
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -150,6 +150,18 @@ in `form_with`.
 
 `form_with` has some important differences compared to `form_for` and `form_tag`, and these differences apply to `bootstrap_form_with`. A good summary of the differences can be found at: https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a, or in the [Rails documentation](api.rubyonrails.org).
 
+
+## Configuration
+
+`bootstrap_form` has some default options that can be overriden:
+
+```ruby
+# config/initializers/bootstrap_form.rb
+BootstrapForm.configure do |c|
+  # c.default_form_attributes = { role: "form" }
+end
+```
+
 ## Form Helpers
 
 `bootstrap_form` provides its own version of the following Rails form helpers:
--- a/bootstrap_form.gemspec
+++ b/bootstrap_form.gemspec
@@ -15,6 +15,8 @@ Gem::Specification.new do |s|
                   "easy to create beautiful-looking forms using Bootstrap 4"
   s.license     = "MIT"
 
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -150,6 +150,18 @@ in `form_with`.
 
 `form_with` has some important differences compared to `form_for` and `form_tag`, and these differences apply to `bootstrap_form_with`. A good summary of the differences can be found at: https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a, or in the [Rails documentation](api.rubyonrails.org).
 
+
+## Configuration
```
lcreid, line 154: I'd like to give a little more information to the reader around configuration. I suggest something like a first paragraph that says:

> `bootstrap_form` can be used out-of-the-box without any configuration. However, `bootstrap_form` does have an optional configuration file at `config/initializers/bootstrap_form.rb` for setting options that affect all generated forms in an application.
>
> 
> The current configuration options are:
> 
> * `default_form_attributes` `bootstrap_form` versions 4.4.0 and older added a `role="form"` attribute to all forms. The W3C validator will raise a warning on forms with a `role="form"` attribute. 

The actual example will depend on the path we choose for maintaining compatibility for version 4, and enabling the new default for version 5 (which will be the Bootstrap v5 version). See my comments on the pull request.

sharshenov:
@lcreid 
> So if I have an application using bootstrap_form version 4.4.0 (the current released version), and something in my app that depends on forms having role="form", will the app break if I just bundle update to version 4.5.0 (if it includes this PR)?

No. Default configuration still makes `role="form"` attribute be rendered. So this change is not breaking. It allows to set `default_form_attributes` with custom hash (like `{}`)

> If so, we need to think of the most practical way to maintain version 4 behaviour, and then make no role="form" the default for version 5 (the Bootstrap 5 version).

I think, the best way is to declare a Github milestone and create an issue to drop `role: "form"` from `BootstrapForm::Configuration::DEFAULT` and then link the issue to the milestone.
I could also add a [post_install_message](https://guides.rubygems.org/specification-reference/#post_install_message) to announce this breaking change in upcoming major version. WDTY? Should I?

> I'd like to give a little more information to the reader around configuration

Updated in https://github.com/bootstrap-ruby/bootstrap_form/pull/562/commits/ce7bc3d64d2eb71603dffadf2f5e100a9177a933 (I'll squash to first commit if it is approved)

lib/bootstrap_form/form_builder.rb
```diff
@@ -58,14 +58,14 @@ def initialize(object_name, object, template, options)
...
+      add_default_form_attributes_and_form_inline options
       super
     end
     # rubocop:enable Metrics/AbcSize
 
-    def add_form_role_and_form_inline(options)
+    def add_default_form_attributes_and_form_inline(options)
       options[:html] ||= {}
-      options[:html][:role] ||= "form"
+      options[:html].reverse_merge!(BootstrapForm.config.default_form_attributes)
```
lcreid, line 68: It would be nice to make this work reasonably if the user sets `default_form_attributes = nil`. Extra bonus marks if we nicely handle if the user sets `default_form_attributes` to something other than a hash. Throwing an error on start-up might be better than waiting until the app actually tries to render a form.

sharshenov:
@lcreid Now configuration raises error if unsupported value is given for `default_form_attributes`. I've also added post install message.
I'd like to squash all commits if you consider PR good enough.

P.S. Travis fails because of unrelated problem with rubocop configuration

bootstrap_form.gemspec
```diff
@@ -15,6 +15,8 @@ Gem::Specification.new do |s|
                   "easy to create beautiful-looking forms using Bootstrap 4"
   s.license     = "MIT"
 
+  s.post_install_message = "Default form attribute role=\"form\" will be dropped in 5.0.0"
```
lcreid, line 18: :+1:

lib/bootstrap_form/configuration.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module BootstrapForm
```
lcreid, line 3: Nice.

PR #565, opened 2020-06-12
consider human attribute name for labels

koenpunt:
I'm not sure if it's like this intentionally, but the `human_attribute_name` was only used when there are errors.


donv:
@koenpunt Do you have any examples where this change makes a difference?  If not, I will close this PR.

koenpunt:
It's been some time, but what I remember is that labels showed as the name of the label, so some for example "first_name" instead of the more human friendly "First name".

donv:
@koenpunt Could you add a test case for this that shows the problem?

UweKubosch:
@koenpunt Thanks for the PR!  Could you add a test case that demonstrates the problem and rebase against main?

Issue #553, opened 2019-12-08
Bootstrap 5


lcreid:
This is a high-level issue to capture the tasks needed for version 5 of `bootstrap_form`/`bootstrap_form` for Bootstrap 5.

Some links to relevant resources:

* https://v5.getbootstrap.com/docs/5.0/migration/#forms
* https://blog.templatetoaster.com/news/bootstrap-5/

To contribute, create an issue or pull request (PR). If you're making a PR, be sure to do it against the `bootstrap-5` branch.

Remember to add to `UPGRADE-5.0.md` if your PR changes the behaviour of the gem, and particularly if the rendered HTML and/or CSS change in a way that might cause the page to render differently.

taylorthurlow:
The Bootstrap 5 alpha [is out today](https://blog.getbootstrap.com/2020/06/16/bootstrap-5-alpha/)!

o-t-w:
The Bootstrap 5 stable release was back in May. Is there any progress on this? Bootstrap 5 looks a lot better and I am desperate to upgrade.

thimo:
A 5.0 release would be quite nice, but in the meantime you can use the bootstrap-5 branch as documented here: https://github.com/bootstrap-ruby/bootstrap_form/tree/bootstrap-5#installation

michaelbaudino:
Shouldn't this be closed, since it has been done in #611 and released as [5.0.0](https://github.com/bootstrap-ruby/bootstrap_form/releases/tag/v5.0.0)?

lcreid:
Thanks, @michaelbaudino . My bad. We appreciate the feedback.

PR #567, opened 2020-06-17
Preliminary support for Bootstrap 5

taylorthurlow:
With the Bootstrap 5 alpha out today, I thought I'd see how hard it would be to update the gem to support the latest alpha release.

This is a work-in-progress PR, but it does seem to be working, at least with the demo application forms. More work needs to be put into updating parts of the README.

I've also questionably removed some tests that I should probably look into figuring out if they're actually tests that should be removed, or slightly reworked.

One thing that might be worth discussing is Bootstrap 5's removal of the `.form-group` class. The concept is pretty hard-wired into the gem. Even though the class is gone, the "concept" of the form group wrapper is still there, just replaced with the `.mb-3` class to maintain the margin on the div. I don't think it would necessarily be a bad thing to keep using the form group terminology within the code base.

I haven't contributed here before, but any existing maintainers should feel free to push to this branch with changes you feel are necessary. If you'd like to go for a more methodical, less "shotgun search and replace" methodology, I wouldn't be offended if my efforts here were scrapped.

### DIFF
--- a/lib/bootstrap_form/form_group.rb
+++ b/lib/bootstrap_form/form_group.rb
@@ -50,7 +50,7 @@ module BootstrapForm
     end
 
     def form_group_classes(options)
-      classes = ["form-group", options[:class].try(:split)].flatten.compact
+      classes = ["mb-3", options[:class].try(:split)].flatten.compact
       classes << "row" if group_layout_horizontal?(options[:layout]) && classes.exclude?("form-row")
       classes << "form-inline" if field_inline_override?(options[:layout])
       classes << feedback_class if options[:icon]
--- a/lib/bootstrap_form/form_group.rb
+++ b/lib/bootstrap_form/form_group.rb
@@ -24,7 +24,7 @@ module BootstrapForm
 
     def form_group_content_tag(name, field_name, without_field_name, options, html_options)
       html_class = control_specific_class(field_name)
-      html_class = "#{html_class} form-inline" if @layout == :horizontal && options[:skip_inline].blank?
+      html_class = "#{html_class} col-auto g-3" if @layout == :horizontal && options[:skip_inline].blank?
       tag.div(class: html_class) do
         input_with_error(name) do
           send(without_field_name, name, options, html_options)
@@ -52,7 +52,7 @@ module BootstrapForm
     def form_group_classes(options)
       classes = ["mb-3", options[:class].try(:split)].flatten.compact
       classes << "row" if horizontal_group_with_gutters?(options[:layout], classes)
-      classes << "form-inline" if field_inline_override?(options[:layout])
+      classes << "col-auto g-3" if field_inline_override?(options[:layout])
       classes << feedback_class if options[:icon]
       classes
     end
--- a/lib/bootstrap_form/form_group.rb
+++ b/lib/bootstrap_form/form_group.rb
@@ -66,7 +66,7 @@ module BootstrapForm
     end
 
     def classes_include_gutters?(classes)
-      classes.any? { |c| c =~ /g-\d+/ }
+      classes.any? { |c| c =~ /^g-\d+$/ }
     end
   end
 end
--- a/lib/bootstrap_form/form_group.rb
+++ b/lib/bootstrap_form/form_group.rb
@@ -51,14 +51,22 @@ module BootstrapForm
 
     def form_group_classes(options)
       classes = ["mb-3", options[:class].try(:split)].flatten.compact
-      classes << "row" if group_layout_horizontal?(options[:layout]) && classes.exclude?("form-row")
+      classes << "row" if horizontal_group_with_gutters?(options[:layout], classes)
       classes << "form-inline" if field_inline_override?(options[:layout])
       classes << feedback_class if options[:icon]
       classes
     end
 
+    def horizontal_group_with_gutters?(layout, classes)
+      group_layout_horizontal?(layout) && !classes_include_gutters?(classes)
+    end
+
     def group_layout_horizontal?(layout)
       get_group_layout(layout) == :horizontal
+    end
+
+    def classes_include_gutters?(classes)
+      classes.any? { |c| c =~ /g-\d+/ }
     end
   end
 end
--- a/lib/bootstrap_form/inputs/base.rb
+++ b/lib/bootstrap_form/inputs/base.rb
@@ -22,6 +22,7 @@ module BootstrapForm
           with_field_name = "#{field_name}_with_bootstrap"
           without_field_name = "#{field_name}_without_bootstrap"
           define_method(with_field_name) do |name, options={}, html_options={}|
+            html_options = html_options.reverse_merge(control_class: "form-select")
             form_group_builder(name, options, html_options) do
               form_group_content_tag(name, field_name, without_field_name, options, html_options)
             end
--- a/lib/bootstrap_form/inputs/check_box.rb
+++ b/lib/bootstrap_form/inputs/check_box.rb
@@ -57,7 +57,7 @@ module BootstrapForm
       end
 
       def check_box_label_class(options)
-        classes = %w[form-label form-check-label]
+        classes = ["form-check-label"]
         classes << options[:label_class]
         classes << hide_class if options[:hide_label]
         classes.flatten.compact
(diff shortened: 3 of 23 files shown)


lcreid:
I've created a `bootstrap-5` branch, and made a few changes to the `README.md`.

Thanks again for your PR!

taylorthurlow:
Switched the base branch. Glad I could be of some help!

taylorthurlow:
I noticed that the `bootstrap-5` branch is based on a commit ~8 commits behind the current master - was this intentional? I just rebased my branch onto `bootstrap-5`, it conflicted on a few really trivial things, but I wanted to double check. Cheers!

thimo:
Thank you so much for picking up support for Bootstrap 5! I'm already using this in an application I'm working on. 

One thing I noticed is that checkbox labels both get the `form-label`and the `form-check-label` classes. Reading the Bootstrap 5 docs at https://v5.getbootstrap.com/docs/5.0/forms/checks/ it looks like the checkbox labels should not use `form-label`, but only `form-check-label`.

thimo:
Another issue: dropdowns (select) should get the class `form-select` instead of `form-control`.

lcreid:
Terribly sorry about that. My bad. I have updated the `bootstrap-5` branch. I hope I didn't cause you too much extra work, @taylorthurlow .

@thimo Thanks for the feedback. We're happy to consider pull requests on the `bootstrap-5` branch if you feel up to it. If it's not too much trouble, could you please create two new issues for your two comments, and reference issue #553 in them so we keep track. 

The reason for separate issues is the following: The upgrade from Bootstrap 3 to 4 was a long process, and some of the individual changes turned out to require significant work. It's easier to track the work for each issue separately, rather than try to do one big PR.

Thanks in advance.

taylorthurlow:
> Terribly sorry about that. My bad. I have updated the `bootstrap-5` branch. I hope I didn't cause you too much extra work, @taylorthurlow .

No problem at all, it took less than 5 minutes. I kept the old branch alive so I just re-pushed the old branch without the small changes.

I'm not sure what the best approach for getting @thimo's proposed fixes merged here. Would it make sense to merge this PR with its flaws intact, and just continue submitting PRs to the `bootstrap-5` branch? Otherwise I could allow @thimo to push to my branch. Whatever you think the best approach is. I'll mark this PR as ready for review so we can merge it if that's the way to go.

lib/bootstrap_form/form_group.rb
```diff
@@ -50,15 +50,23 @@ def form_group_control_class(options)
...
+    def horizontal_group_with_gutters?(layout, classes)
+      group_layout_horizontal?(layout) && !classes_include_gutters?(classes)
+    end
+
     def group_layout_horizontal?(layout)
       get_group_layout(layout) == :horizontal
     end
+
+    def classes_include_gutters?(classes)
+      classes.any? { |c| c =~ /g-\d+/ }
```
lcreid, line 69: Do you need to anchor this regex at the start or end of the string? I'm thinking about a case where the user is using a class name like `setting-5`. (I haven't even looked at the Bootstrap 5 classes yet, so please forgive me if this comment has an obvious answer.)

lib/bootstrap_form/inputs/check_box.rb
```diff
@@ -50,40 +50,26 @@ def check_box_value(name, value)
       end
 
       def check_box_classes(name, options)
-        classes = [options[:class]]
-        classes << (options[:custom] ? "custom-control-input" : "form-check-input")
+        classes = [options[:class], "form-check-input"]
```
lcreid, line 53: Your code appears equivalent to the legacy code with respect to `options[:class]`. However, I'm wondering if a user passed in an array of classes, would the code work? In many places the user can provide an array of classes. (I think it's supported by the underlying Rails helpers.) Maybe we should make this work for the case where someone provides an array of class names in `options[:class]`. What do you think?

lib/bootstrap_form/inputs/radio_button.rb
```diff
@@ -34,43 +34,25 @@ def radio_button_label(name, value, options)
       end
 
       def radio_button_classes(name, options)
-        classes = [options[:class]]
-        classes << (options[:custom] ? "custom-control-input" : "form-check-input")
+        classes = [options[:class], "form-check-input"]
```
lcreid, line 37: Same comment as above for check boxes, for `options[:class]` being an array.

test/bootstrap_checkbox_test.rb
```diff
@@ -477,63 +477,6 @@ class BootstrapCheckboxTest < ActionView::TestCase
                                                                     :street, checked: collection)
   end
```
lcreid, line 479: So the whole "custom" thing is gone? That makes life easier.

lcreid:
@taylorthurlow Apologies for the disorganized nature of my feedback. My paid job doesn't use the gem, so I have to squeeze in time around family and work. Anyway, now that I have my head around @thimo 's comments, I'll leave it up to you if you want to add them to this PR. They seem pretty straightforward, so if you don't mind fixing them in this PR, it would move things along faster.

lib/bootstrap_form/form_group.rb
taylorthurlow: Your concern makes total sense - and you're right, it should be anchored - I'd say both sides.

lib/bootstrap_form/inputs/check_box.rb
taylorthurlow: I just confirmed for myself that the `class` option supports an array of strings. That's cool, didn't know that was supported. Good catch, I'll refactor to account for the option.

test/bootstrap_checkbox_test.rb
taylorthurlow: No kidding, much easier. [Here's the relevant section](https://v5.getbootstrap.com/docs/5.0/migration/#forms) of the Bootstrap 5 migration docs. They're clearly incomplete and missing significant parts, but the linked section says about as much - the custom stuff is gone.

taylorthurlow:
@lcreid Not disorganized at all, thanks for the review. I'll fix the things that @thimo has mentioned tomorrow. 👍

lib/bootstrap_form/inputs/check_box.rb
taylorthurlow: I'm realizing now that this is why `#flatten` is called in the return value - so this already worked. That said, I'd rather this be clearer so I'll fix it nonetheless.

taylorthurlow:
I've fixed the issues @thimo brought up. The `form-check-label` issue was a quick fix, but the `form-control` to `form-select` was not.

I feel like the way I fixed it is a little sloppy. The control class is `form-control` by default, and because this was the correct control class for the select box in 4.x, all select fields use the control class definition from `BootstrapForm::FormBuilder#control_class`. In order to override this I copied the `#reverse_merge` technique found here:

https://github.com/taylorthurlow/bootstrap_form/blob/bootstrap5/lib/bootstrap_form/inputs/file_field.rb#L11

And I applied that to all necessary select fields, but for `html_options` instead.

Let me know if you have any suggestions on how to make this a little better.

lib/bootstrap_form/inputs/base.rb
```diff
@@ -22,6 +22,7 @@ def bootstrap_select_group(field_name)
           with_field_name = "#{field_name}_with_bootstrap"
           without_field_name = "#{field_name}_without_bootstrap"
           define_method(with_field_name) do |name, options={}, html_options={}|
+            html_options = html_options.reverse_merge(control_class: "form-select")
```
lcreid, line 25: I spent some time looking at this, refreshing my memory about how the code works. The `form_group_builder` method made more sense in earlier versions of Bootstrap, I think. I think your solution is good. I've created an issue #568 to look at whether `form_group_builder` is really serving a purpose now. But your PR is good.

taylorthurlow: Good to hear. The metaprogramming involved made it a little harder to understand what was happening. 👍

thimo:
Awesome work, thanks a lot! I've switched to using the new `bootstrap-5` branch and will report back if I run into anything.

Issue #579, opened 2021-03-24
Errors on helper uses weird class names


cseelus:
The [`errors_on` helper](https://github.com/bootstrap-ruby/bootstrap_form/blob/4d92a8af1dd7b176b0b8e3151c600ce2607722d4/lib/bootstrap_form/helpers/bootstrap.rb#L45) outputs a div with `alert` and `alert-danger` classes, [typically used](https://www.google.com/search?client=safari&rls=en&q=rails+flash+bootstrap&ie=UTF-8&oe=UTF-8) for Rails flash messages.

<img width="674" alt="Screen Shot 2021-03-24 at 21 14 35" src="https://user-images.githubusercontent.com/3188392/112384134-f3c36380-8ce5-11eb-8e0a-df9d8ac9930c.png">

Why not use the `invalid-feedback` class name, like with other form errors?

Could do a PR, if there is a consensus on this change, alternatively add an option to change those class names.

lcreid:
Thanks for taking the time to create an issue. You make an good point. I don't use `errors_on` myself, so I hadn't really considered it.

On of the challenges we have is maintaining backwards compatibility for all the existing applications that use this gem. I'd be happy to take a PR, but I'd like to hear how we can use the consisten classes, without breaking existing applications. Perhaps a configuration option, but I'll leave the thinking up to you. 

I suggest you throw some questions of ideas in a reply here so we can discuss them before you do a PR, but I'm sure we can some up with something so you can prepare a PR.

Thanks again for your suggestion and interest in `bootstrap_form`.

cseelus:
To keep it backwards compatible (within the current major version?), yes I too think the only way is to add another configuration option (besides the `hide_attribute_name` one). 

My suggestion would be to call it `inline` (e.g. `inline: true`) in reference to the default inline error message.

So to get the default inline error output, one would have to use it like this:

```erb
<%= f.errors_on :tasks, hide_attribute_name: true, inline: true %>
```

Calling that option `inline` might be a bit redundant, as the docs state this helper is for *custom inline errors*, so another option would be to call it `feedback`, in reference to the default inline errors.

lcreid:
Thanks for your patience with my slow reply. And thanks for the suggestions. I'm now thinking that we probably just completely overlooked the `errors_on` method when Bootstrap 4 changed its error classes. I still don't think we should break Bootstrap 4 applications that use `bootstrap_form`, but we should make the default consistent for Bootstrap 5.

[...]

PR #580, opened 2021-04-03
Errors on invalid feedback

cseelus:
Follow up on [issue 579](https://github.com/bootstrap-ruby/bootstrap_form/issues/579).

Took a little while, since I first wanted the project we use `bootstrap_form` with, to use Bootstrap 5 in a new development branch. So far the transition went smoother than the upgrade from 3 to 4, although it is far from complete.

The `bootstrap-5` branch of `bootstrap_form` works surprisingly well already.


lcreid (commented):
This looks good so far. Could you please add some documentation in `UPGRADE-5.0.md` to describe this change and perhaps tell people what they would have to do if they can't or don't want to change their CSS from the previous version? Thanks in advance.

cseelus:
I would add the following text to the bottom of `UPGRADE-5.0.md`:

---

## Different behavior for `errors_on` helper

The `errors_on` helper now wraps the error message in a CSS class `invalid-feedback`, instead of `alert` and `alert-danger`, as before.

This will display the error as any other [Bootstrap inline form error](https://getbootstrap.com/docs/5.0/forms/validation/#server-side), instead of displaying it as an [Bootstrap alert](https://getbootstrap.com/docs/5.0/components/alerts/).

---

Could you elaborate on what you mean by "tell people what they would have to do if they can't or don't want to change their CSS from the previous version"? Then I will add this also.
As far as I understand it, this should work without any extra CSS (as with < 5.0).

lcreid:
> Could you elaborate on what you mean by "tell people what they would have to do if they can't or don't want to change their CSS from the previous version"? Then I will add this also.
> As far as I understand it, this should work without any extra CSS (as with < 5.0).

I'm thinking of the (unlikely) case where someone might have deliberately styled the CSS classes differently. Then they would have to do something to get the same results as before. And if they can't do anything about it, perhaps they should be able to. Maybe we could add a `class` option to `errors_on`, where they can specify a class or classes that override the default `invalid-feedback`. 

It may seem like an unlikely case, but I think it's good practice in general to give people a way to keep using our gem,. Another solution might be to document that if the new behaviour of `errors_on` is unacceptable, they can hard-code the HTML and CSS in place of using `errors_on`, or write their own method that replaces it. But I think adding a class argument to the method (and documenting it) feels like not too much effort. Let me know if I'm wrong.

What you wrote above is just perfect for what I originally asked for.

Thanks again for taking the time to contribute to `bootstrap_form`.

cseelus:
So you mean an option like `class: 'foo'`, that will just override '.invalid-feedback'?

Regarding the custom helper idea you asked for, I played around with the idea for a bit:

```ruby
  def custom_errors_on(form, name, options = {})
    return if form.object.errors[name].blank?

    tag.div(class: 'alert alert-danger') do
      # hidden input tag is needed for the 'invalid-feedback' to be shown
      concat tag.input(type: 'hidden', class: 'is-invalid')
      concat form.errors_on(name, options)
    end
  end
```

lcreid:
> So you mean an option like `class: 'foo'`, that will just override '.invalid-feedback'?

Yeah, basically something like that. BTW, I hope I'm not frustrating you, but I really don't have the cycles to think through the details, so I like to year what you propose. But, yeah, `class: 'foo'` is probably all we need here. (And see my comment at the end.)
> 
> Regarding the custom helper idea you asked for, I played around with the idea for a bit:
> 
> ```ruby
>   def custom_errors_on(form, name, options = {})
>     return if form.object.errors[name].blank?
> 
>     tag.div(class: 'alert alert-danger') do
>       # hidden input tag is needed for the 'invalid-feedback' to be shown
>       concat tag.input(type: 'hidden', class: 'is-invalid')
>       concat form.errors_on(name, options)
>     end
>   end
> ```

Thanks for doing the work to play with this. It seems to me like adding the `class` option to `errors_on`, plus documentation, isn't very much work compared to having to tell people how to write their own method to override ours. Maybe we could just add the `class` option? What do you think?

cseelus:
I've added a custom class option for this helper.

lib/bootstrap_form/helpers/bootstrap.rb
```diff
@@ -46,8 +46,9 @@ def errors_on(name, options={})
         return unless error?(name)
 
         hide_attribute_name = options[:hide_attribute_name] || false
+        custom_class_name   = options[:custom_class_name] || false
```
lcreid, line 49: I think the `_name` isn't really needed here, but that's not a strong opinion, so if you prefer to leave it, please do so.

UPGRADE-5.0.md
```diff
@@ -11,3 +11,9 @@ Upgrading `bootstrap_form` to version 5 means you must upgrade your whole applic
 ## No `role="form"` Attribute
 
 As explained in #560, the `role="form"` attribute generated by `bootstrap_4` caused the W3C validator to output a warning. The `role="form"` attribute was deprecated in the 4.5.0 and is being remove completely in 5.0.0. This has no impact on `bootstrap_form` code itself, but may affect your application if it depended on a form having this attribute set. (Issue #569)
+
+## Different behavior for `errors_on` helper
+
+The `errors_on` helper now wraps the error message in a CSS class `invalid-feedback`, instead of `alert` and `alert-danger`, as before.
+
+This will display the error as any other [Bootstrap inline form error](https://getbootstrap.com/docs/5.0/forms/validation/#server-side), instead of displaying it as an [Bootstrap alert](https://getbootstrap.com/docs/5.0/components/alerts/).
```
lcreid, line 19: What do you think about adding a short note here to tell people what to do if they need the old behaviour? In other words, tell them about the option they can now provide to `errors_on` to change the class name(s). 

If you add a description to the README, you could even just redirect them from here to the relevant documentation in the README.

lcreid (commented):
Thanks. This looks good. 

Could you please add a simple test case to ensure the option generates the HTML you expect? Your change is simple and I'm sure it works, but we need the test to ensure that we don't break the behaviour in the future if, for example, we do some major refactoring or clean-up of the code.

We need to tell people about the new option in the README. There's a short section for `errors_on`. Just mention the option and perhaps give a quick example (maybe grab the test case).

Thanks again for taking the time to make this PR. We're almost there.

lib/bootstrap_form/helpers/bootstrap.rb
cseelus: Yes, its better without `_name`

UPGRADE-5.0.md
cseelus: This should indeed be documented as well.

cseelus:
Now that the approach seems to be ok, I added tests and more info about this change to the docs and into the changelog.

README.md
```diff
@@ -803,7 +802,19 @@ You can hide the attribute name like this:
 Which outputs:
 
 ```html
-<div class="alert alert-danger">can't be blank.</div>
+<div class="invalid-feedback">can't be blank.</div>
```
lcreid, line 805: Nice. Clear and to the point. I like it.

UPGRADE-5.0.md
```diff
@@ -11,3 +11,15 @@ Upgrading `bootstrap_form` to version 5 means you must upgrade your whole applic
...
+
+You can use the `custom_class` options for this helper with `alert alert-danger` to restore the old behaviour:
```
lcreid, line 21: Again. Very clear.

test/bootstrap_form_test.rb
```diff
@@ -793,8 +793,17 @@ class BootstrapFormTest < ActionView::TestCase
     @user.email = nil
     assert @user.invalid?
 
-    expected = '<div class="alert alert-danger">can&#39;t be blank, is too short (minimum is 5 characters)</div>'
+    expected = '<div class="invalid-feedback">can&#39;t be blank, is too short (minimum is 5 characters)</div>'
 
     assert_equivalent_xml expected, @builder.errors_on(:email, hide_attribute_name: true)
   end
+
+  test "errors_on use custom CSS classes" do
```
lcreid, line 801: :+1:

lcreid (approved):
Looks great! I'll resolve the merge conflict and merge it. Thanks again for all your work on this.

PR #583, opened 2021-04-06
Fix tests in bootstrap-5 branch. Fixes #582

thimo:
Fixes the rendering tests in the bootstrap-5 branch after Rails rendering changes.

Also includes changes to remove Rubocop offences.

PR #584, opened 2021-04-07
Support Bootstrap 5 floating labels. Fixes #573

### LINKED ISSUE #573: Support Bootstrap 5 alpha 3 floating labels
Bootstrap 5 alpha 3 supports floating labels: https://v5.getbootstrap.com/docs/5.0/forms/floating-labels/

It would be nice to enable this in bootstrap_form with an option, something like `<%= f.text_field :query, floating: true %>`.

To enable this a few things must be done in the rendering:

1. use `form-floating` as the control class
2. revert order of `input` and `label` elements
3. add `placeholder` to input field, fill with label value by default

```
<div class="form-floating mb-3">
  <input placeholder="Search" class="form-control" type="text" name="filter[query]" id="filter_query">
  <label for="filter_query">Search</label>
</div>
```

I would love to pick this up, but if anybody else has the time, please feel free to do so. Would also love feedback on how to enable this functionality, is `floating: true` the right way to do that?

thimo:
Add support for Bootstrap 5 floating labels. See https://github.com/bootstrap-ruby/bootstrap_form/issues/573.

CleanShot 2021-04-07 at 10 20 48@2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -693,6 +693,21 @@ The `custom` option can be used to replace the browser default styles for check 
 <% end %>
 ```
 
+### Floating labels
+
+The `floating` option can be used to enable Bootstrap 5's floating labels. This option is supported on text fields
+and dropdowns. Here's an example:
+
+```erb
+<%= bootstrap_form_for(@user) do |f| %>
+  <%= f.email_field :email, floating: true %>
+  <%= f.password_field :password, floating: true %>
+  <%= f.password_field :password, floating: true %>
+  <%= f.select :status, [["Active", 1], ["Inactive", 2]], include_blank: "Select a value", floating: true %>
+  <%= f.submit "Log In" %>
+<% end %>
+```
+
 ## Validation and Errors
 
 Rails normally wraps fields with validation errors in a `div.field_with_errors`, but this behaviour isn't consistent with Bootstrap 4 styling. By default, `bootstrap_form` generations in-line errors which appear below the field. But it can also generate errors on the label, or not display any errors, leaving it up to you.
--- a/README.md
+++ b/README.md
@@ -693,7 +693,7 @@ The `custom` option can be used to replace the browser default styles for check 
 <% end %>
 ```
 
-### Floating labels
+### Floating Labels
 
 The `floating` option can be used to enable Bootstrap 5's floating labels. This option is supported on text fields
 and dropdowns. Here's an example:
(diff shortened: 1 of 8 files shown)


thimo:
Tests fail, but this should be fixed when merging https://github.com/bootstrap-ruby/bootstrap_form/pull/583.

test/bootstrap_fields_test.rb
```diff
@@ -428,4 +428,14 @@ class BootstrapFieldsTest < ActionView::TestCase
       end
     end
   end
+
+  test "can have a floating label" do
+    expected = <<-HTML.strip_heredoc
```
lcreid, line 433: We can use `<<~` now, instead of `strip_heredoc`. We no longer support Ruby versions that _don't_ have `<<~`.

lcreid (commented):
I really appreciate all the work you've done. I'd just ask that you add something in the README (ideally with example) showing how to make floating labels with `bootstrap_form`. Thanks in advance.

test/bootstrap_fields_test.rb
thimo: Sounds great, I wasn't really aware of this (just copy/pasted an existing test). But this would be the first and only occurrence of `<<~` in the project where `strip_heredoc` is used 220 times. Mind if I change all of them in this PR? Or would you prefer to only have this one changed and create a new PR for the rest?

thimo:
I've added a section "Floating Labels" to the "Form Styles", including a short example. Nothing too fancy. Please let me know if this can/should be improved on.

Additionally I've rebased my branch to the current `bootstrap-5` branch so tests would run, and also fixed one Rubocop offence.

test/bootstrap_fields_test.rb
thimo: Never mind, just saw you added the PR https://github.com/bootstrap-ruby/bootstrap_form/issues/585. Will pick that one up after this and change this single occurrence of `strip_heredoc`.

README.md
```diff
@@ -693,6 +693,21 @@ The `custom` option can be used to replace the browser default styles for check
 <% end %>
 ```
 
+### Floating Labels
```
lcreid, line 696: :fire: This is fine. Thanks!

lcreid (approved):
Nice extra test coverage. This looks great. Thanks again!

PR #602, opened 2021-08-09
Get bot access token from secrets

lcreid:
Just try to run CI.


donv:
I changed to use the default github.token for danger instead.  Closing this PR.

PR #603, opened 2021-09-21
fix: Auto-correct RuboCop issues

olleolleolle:
This PR fixes the remaining **linting** issues.

### DIFF
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -2,7 +2,10 @@ module BootstrapForm
   module Helpers
     module Bootstrap
       def button(value=nil, options={}, &block)
-        value, options = nil, value if value.is_a?(Hash)
+        if value.is_a?(Hash)
+          options = value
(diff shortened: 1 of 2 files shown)


olleolleolle:
There was a Faraday deprecation notice, but also an octokit auth failure. Perhaps I do not have the correct credentials?

```
WARNING: `Faraday::Connection#authorization` is deprecated; it will be removed in version 2.0.
While initializing your connection, use `#request(:authorization, ...)` instead.
See https://lostisland.github.io/faraday/middleware/authentication for more usage info.
```

lcreid:
Yes. I just haven't had the time or motivation to figure out how to run Danger through GitHub actions. It's fairly well documented, but it wasn't completely clear to me how to store the credentials, so I haven't done it yet. If you want to take a run at it, I'd be happy to review the PR.

lcreid (approved):
Thanks for this. I believe in using tools like RuboCop (obviously), but the constant changes to the "standards" can be a bit tiring to keep up with.

PR #613, opened 2021-11-17
Fix Rails Edge tests

### LINKED ISSUE #609: Test Fail on Edge AKA Rails 7
Tests are currently failing on edge Rails, AKA Rail 7. If anyone is interested in working on a fix, please comment below.

Please branch from and submit pull requests to the ~`bootstrap-5`~ `main` branch. Thanks in advance.

thimo:
Tests would no longer even start to run because `Rails.application.config.assets` was not available to set in `demo/config/initializers/assets.rb`. Adding gem `sassc-rails` makes the part of config available again.

Next rendering changed for hidden fields ('autocomplete="off"' was added) and rich text area.

Additionally a few Rubocop offences have been removed.

Fixes #609.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,5 +21,6 @@ group :test do
   gem "diffy"
   gem "equivalent-xml"
   gem "mocha"
+  gem "sassc-rails"
   gem "sqlite3"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gemspec path: __dir__
 # gem "rails", "~> 5.2.0"
 # gem "rails", "~> 6.0.0"
 # gem "rails", "~> 6.1.0"
-# gem "rails", git: "https://github.com/rails/rails.git"
+# gem "rails", git: "https://github.com/rails/rails.git", branch: "main"
 
 group :development do
   gem "chandler", ">= 0.7.0"
--- a/bootstrap_form.gemspec
+++ b/bootstrap_form.gemspec
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.description = "bootstrap_form is a rails form builder that makes it super "\
                   "easy to create beautiful-looking forms using Bootstrap 5"
   s.license     = "MIT"
+  s.metadata    = { "rubygems_mfa_required" => "true" }
 
   s.files = `git ls-files -z`.split("\x0").reject do |f|
     f.match(%r{^(test)/})
--- a/test/bootstrap_rich_text_area_test.rb
+++ b/test/bootstrap_rich_text_area_test.rb
@@ -1,4 +1,5 @@
 require_relative "./test_helper"
+require "minitest/mock"
 
 if ::Rails::VERSION::STRING > "6"
   class BootstrapRichTextAreaTest < ActionView::TestCase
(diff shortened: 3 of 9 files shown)


test/bootstrap_rich_text_area_test.rb
```diff
@@ -8,18 +9,39 @@ class BootstrapRichTextAreaTest < ActionView::TestCase
     setup :setup_test_fixture
 
     test "rich text areas are wrapped correctly" do
-      expected = <<~HTML
-        <div class="mb-3">
-          <label class="form-label" for="user_life_story">Life story</label>
-          <input type="hidden" name="user[life_story]" id="user_life_story_trix_input_user"/>
+      with_stub_token do
```
lcreid, line 12: I must confess I'm too tired tonight to get my head around this test. I'll take another look tomorrow. Sorry.

lcreid (commented):
Great stuff. I'm glad you figured these things out. I'm still a little curious about why adding `sassc-rails` fixes the config issue. Do you think we should be doing something else about the configuration?

I'll try to have more useful feedback tomorrow night.

bootstrap_form.gemspec
```diff
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.description = "bootstrap_form is a rails form builder that makes it super "\
                   "easy to create beautiful-looking forms using Bootstrap 5"
   s.license     = "MIT"
+  s.metadata    = { "rubygems_mfa_required" => "true" }
```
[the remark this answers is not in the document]
thimo: Oh yeah, I probably should have mentioned this. 😬 It's something that Rubocop wanted us to add. Maybe, if we can't do this, we can put a "disable warning" block here. But on the other hand, I think the goal here is to make the rubygems login more secure to prevent hacks.

test/bootstrap_rich_text_area_test.rb
thimo: The `with_stub_token` is copied from a test in the Rails project itself (https://github.com/rails/rails/blob/main/actiontext/test/template/form_helper_test.rb) and makes sure that `data-direct-upload-token` is always the same value "token", so that we can validate the HTML output. Without this stub the value for `data-direct-upload-token` would be a big, random string.

On top of that, the markup for trim in Rails 7 differs in quite a few areas so I thought it would make sense to just make a big IF statement instead of doing everything inline and making a big(ger) mess.

Hope this makes sense. 😀

Gemfile
```diff
@@ -21,5 +21,6 @@ group :test do
   gem "diffy"
   gem "equivalent-xml"
   gem "mocha"
+  gem "sassc-rails"
```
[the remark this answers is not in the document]
thimo: I agree that it's curious, but I'm not sure about the best approach here. Regarding adding `sassc-rails`, it was already included in the `development` group, so I didn't see too much harm in adding it to `test`. I would assume it has no impact of users of `bootstrap_form` for this reason, but that's just an assumption. 

Another way to solve/work around this issue is to remove `demo/config/initializers/assets.rb` or comment-out its two remaining lines. I'm mostly concerned for the `Rails.application.config.assets.paths << Rails.root.join("node_modules")` line, but we may not need that one.

bootstrap_form.gemspec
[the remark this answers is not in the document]
lcreid: I'm all for better security. When I finally pushed the 5.0 version of the gem I was wondering if I was going to have to give 2FA, but it just worked so I didn't question it. Glad you followed up.

test/bootstrap_rich_text_area_test.rb
lcreid: Thanks for the explanation. It makes sense. See my comment a little further down.

test/bootstrap_rich_text_area_test.rb
```diff
@@ -8,18 +9,39 @@ class BootstrapRichTextAreaTest < ActionView::TestCase
...
+        HTML
+      end
     end
 
     def data_blob_url_template
       "http://test.host/rails/active_storage/blobs/#{'redirect/' if ::Rails::VERSION::STRING >= '6.1'}:signed_id/:filename"
     end
+
+    def with_stub_token(&block)
+      return if ::Rails::VERSION::STRING < "7"
```
lcreid, line 42: I think this means you skip running the assertion, which I don't think you intended. Maybe you want to just `yield` if the Rails version is less than 7?

lcreid (commented):
Thanks for helping my old brain along. I think there's one thing you might want to change, if I'm not missing something. See the comments.

test/bootstrap_rich_text_area_test.rb
thimo: You're right, thanks for catching this! `return yield` would fix it, but looking at the test as a whole again all the checks for Rails 7 make it less readable. So I now choose to have one bit of mark-up and test for >= Rails 7 and one for older versions. Makes it a bit less DRY, but I think this will make it easier to understand in the future.

test/bootstrap_rich_text_area_test.rb
```diff
@@ -8,18 +9,35 @@ class BootstrapRichTextAreaTest < ActionView::TestCase
...
     test "rich text areas are wrapped correctly" do
-      expected = <<~HTML
-        <div class="mb-3">
-          <label class="form-label" for="user_life_story">Life story</label>
-          <input type="hidden" name="user[life_story]" id="user_life_story_trix_input_user"/>
-          <trix-editor id="user_life_story" data-blob-url-template="#{data_blob_url_template}" data-direct-upload-url="http://test.host/rails/active_storage/direct_uploads" input="user_life_story_trix_input_user" class="trix-content form-control" />
-        </div>
-      HTML
-      assert_equivalent_xml expected, form_with_builder.rich_text_area(:life_story)
+      if ::Rails::VERSION::STRING >= "7"
```
lcreid, line 12: I'm all for making the code more understandable, if making it DRY conflicts with making it easy to understand. Good call!

lcreid (approved):
Nice work!. Thanks!

PR #634, opened 2022-01-25
Fix text alignment in alert message when there is no error summary

isaac:
When there is no error summary the alert message is wrapped in a p tag. This causes the message to be misaligned vertically. In the Bootstrap examples where there is single line of text, the text is not wrapped in a p tag: https://getbootstrap.com/docs/5.0/components/alerts/#examples

Screenshots below illustrate the issue.

Before:
Before
After:
After

### DIFF
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -27,8 +27,12 @@ module BootstrapForm
         return unless object.respond_to?(:errors) && object.errors.full_messages.any?
 
         tag.div class: css do
-          concat tag.p title
-          concat error_summary unless options[:error_summary] == false
+          if options[:error_summary] == false
+            title
+          else
+            concat tag.p title
+            concat error_summary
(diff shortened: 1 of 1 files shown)


lcreid:
Thanks for catching this, and thanks for the PR. It looks good, but I'd appreciate it if you can fix the test cases so they all pass. Don't worry about the Danger test failing -- there's an issue with Danger on GitHub that doesn't seem to have a solution, but we haven't disabled Danger yet. Sorry.

UweKubosch:
@isaac Thanks for the PR!  Could you fix the one failing test case and rebase against main?

isaac:
@lcreid @UweKubosch @donv 

Sorry for the slow response!

I couldn't find a failing test case - has this already been fixed somewhere else?

donv:
@isaac Yes, I fixed it 😄

donv:
Examples in thre README:  https://github.com/bootstrap-ruby/bootstrap_form#alert-messages

PR #635, opened 2022-04-23
Add example images to README.md

donv:
I enhanced the demo app to accept an ERB snippet and render it.

Then I added a new system test to the demo app that scans the README.md file for ERB snippets, renders them, saves a screenshot, and inserts an image link for the screenshot in the README.md file.

This is both to display the result to developers, but also to ensure that the snippet actually does what it should and track changes over time.

What do you think?


lcreid (approved):
This is brilliant. I love it. Thank you so much for this work. 

A few of the examples seem to render as `...` for me, but we can fix those later.

PR #636, opened 2022-04-24
Add generated HTML for the examples in README.md

donv:
Left the existing HTML examples alone for now.  I want to update them later to see if we are still generating the expected HTML.

Added a border around the example images for better visibility in README.md.

Also removed images for empty examples in README.md


lcreid (approved):
Thanks for continuing with this. I think it looks better with the outline. Feel free to merge this yourself, or let me know and I'll merge it.

PR #638, opened 2022-04-28
Add margin below check boxes in the vertical layout

donv:
For vertical layouts and a single check box, the form group should have a bottom margin.


donv:
Thanks!  I really like how we now can verify the change with a visual example 😄 .

lcreid:
@donv In hindsight, I don't think this was the right decision. There are lots of cases where someone might want invidiual check boxes to be spaced the same as radio buttons. Comgined with [this issue](https://github.com/bootstrap-ruby/bootstrap_form/issues/738), it becomes pretty hard to achieve the desired output.

However, given that this has been out there for three years, I don't want to just change back.

We've started to be able to define some configuration, and I've long wished we had a way to warn about changes to rendered output, and allow configuration based on version. What do you think if I look into ways to configure default behaviour, and warn about changes to defaults?

PR #639, opened 2022-04-29
Update the inline form layout

donv:
Bootstrap 5 uses new classes to style the inline forms.

Reference: https://getbootstrap.com/docs/5.1/forms/layout/#inline-forms


lcreid (commented):
I'm late, but this looks great! Thanks for all your work to bring this gem up-to-date! :100:

PR #646, opened 2022-06-14
Remove dependecies from gemspec

lcreid:
It's not possible to specify the dependencies in the gemspec. `ActionText` only applies if the Rails version is greater than or equal to 6. This gem doesn't make sense unless it's used with Rails anyway, so there's not much value in having dependencies in the gemspec. 

For simple development and testing, the `Gemfile` in the top-level directory pulls in the minimal components of Rails.


lcreid:
#651 is a better solution for this.

PR #647, opened 2022-06-25
More portable `docker-compose.yml` and better documentation

lcreid:
Removed some of the Linux-specific config from `docker-compose.yml`, and improved the documentation showing `docker-compose.override.yml` that exposes the host's credentials in the container, so you can release the gem from within the Docker container.


lcreid:
Requesting review if anyone else is using Docker to maintain this gem.

PR #651, opened 2022-07-20
This allows running tests on the root of the project freshly checked out

donv:
Now `bundle exec rake` works out of the box.


lcreid (approved):
This is awesome! Great find and great work!

PR #652, opened 2022-07-24
Change `concat` to `<<` since the buffer no longer supports the former

donv:


donv:
Woohoo!  Green tests!

lcreid (approved):
Thanks for fixing this. I suspect the test started to fail with https://github.com/rails/rails/pull/45614, which is a PR that speeds up template processing noticeably. Re-reading the documentation, it looks like the use of `concat` was never really encouraged anyway, so it's not surprising that this usage isn't really considered to be something that needs to be supported.

PR #654, opened 2022-08-10
Automatically pick up presence validators for default belongs_to association

donv:
Automatically pick up presence validators for an association for a given ID attribute.

This enables form validation with messages for default `belongs_to` associations since Rails 5.


lib/bootstrap_form/components/validation.rb
```diff
@@ -8,21 +8,23 @@ module Validation
...
-                            else
-                              []
-                            end
+        presence_validator?(target_validators(target, attribute))
+      end
 
-        presence_validator?(target_validators)
+      def target_validators(target, attribute)
+        target_validators = target.validators_on(attribute).map(&:class)
+        target_validators.concat target.validators_on(attribute[0..-4]).map(&:class) if attribute.end_with?("_id")
```
lcreid, line 26: I believe an association can specify any string for the foreign key, so I wonder if just looking for "_id" at the end of the attribute is going to cover all cases. I'm guessing there's a way to access the foreign key column name for an association.

lcreid (commented):
This is a great find, and a nice solution. I'm hoping it's not too hard to make it more general, for when the foreign key column name doesn't match Rails' conventions.

It would be nice to see a test to show what the new code does.

lib/bootstrap_form/components/validation.rb
UweKubosch: Absolutely, but I wanted to cover the most common case first to get the big win for little work 😄 

Can we put it in a separate issue/PR?

UweKubosch:
I'll do at least an entry in the README to see what the code does.

lib/bootstrap_form/components/validation.rb
lcreid: Good plan. Please create the issue and reference it here. I'll approve and merge. Are you going to work on the more generalized solution?

lcreid (approved):
I won't merge this in case you want to do the README in this PR. Otherwise, you can merge this and do another PR.

donv:
Hey good people!  I added the fully generalized version and added a section in the README displaying the resulting code.

Please have a look.  Squash and merge if it looks good.

donv:
Oh, wait!  One more fix needed!

donv:
Now, go!

Issue #663, opened 2022-10-27
Set required and aria on input when input is required


CharlieWinkwaves:
I fixed this before in https://github.com/bootstrap-ruby/bootstrap_form/pull/605 and https://github.com/bootstrap-ruby/bootstrap_form/pull/606 for both bootstrap 4 and 5 but somehow it got overwritten by a merge??

So here are the changes again

donv:
@CharlieWinkwaves Thank you! Where are the changes?

PR #664, opened 2022-10-27
Set required and aria-required on the input when the input is required

CharlieWinkwaves:
https://github.com/bootstrap-ruby/bootstrap_form/issues/663

### DIFF
--- a/lib/bootstrap_form/components/labels.rb
+++ b/lib/bootstrap_form/components/labels.rb
@@ -25,13 +25,7 @@ module BootstrapForm
       def label_classes(name, options, custom_label_col, group_layout)
         classes = ["form-label", options[:class], label_layout_classes(custom_label_col, group_layout)]
 
-        case options.delete(:required)
-        when true
-          classes << "required"
-        when nil, :default
-          classes << "required" if required_attribute?(object, name)
-        end
-
+        classes << "required" if options.delete(:required)
         classes << "text-danger" if label_errors && error?(name)
         classes.flatten.compact
       end
--- a/lib/bootstrap_form/components/labels.rb
+++ b/lib/bootstrap_form/components/labels.rb
@@ -24,8 +24,8 @@ module BootstrapForm
 
       def label_classes(name, options, custom_label_col, group_layout)
         classes = ["form-label", options[:class], label_layout_classes(custom_label_col, group_layout)]
-
-        classes << "required" if options.delete(:required)
+        options = required_field_options(options, name, options)
+        classes << "required" if options.delete(:required) && options[:aria].delete(:required)
         classes << "text-danger" if label_errors && error?(name)
         classes.flatten.compact
       end
--- a/lib/bootstrap_form/components/labels.rb
+++ b/lib/bootstrap_form/components/labels.rb
@@ -24,8 +24,8 @@ module BootstrapForm
 
       def label_classes(name, options, custom_label_col, group_layout)
         classes = ["form-label", options[:class], label_layout_classes(custom_label_col, group_layout)]
-        options = required_field_options(options, name, options)
-        classes << "required" if options.delete(:required) && options[:aria].delete(:required)
+        classes << "required" if required_field_options(options, name)[:required]
+        options.delete(:required)
         classes << "text-danger" if label_errors && error?(name)
         classes.flatten.compact
       end
--- a/lib/bootstrap_form/form_group_builder.rb
+++ b/lib/bootstrap_form/form_group_builder.rb
@@ -28,7 +28,8 @@ module BootstrapForm
     def form_group_builder_options(options, method)
       options.symbolize_keys!
       options = convert_form_tag_options(method, options) if acts_like_form_tag
-      options[:required] = form_group_required(options) if !options[:skip_label] && options.key?(:skip_required)
+      options[:required] = form_group_required(options, method)
+      options[:aria] = { required: true } if options[:required]
       options
     end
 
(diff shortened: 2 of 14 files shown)


donv:
@CharlieWinkwaves Thanks!  Can you fix the two rubocop offences?

lib/bootstrap_form/inputs/check_box.rb
```diff
@@ -26,6 +23,14 @@ def check_box_with_bootstrap(name, options={}, checked_value="1", unchecked_valu
 
       private
 
+      def check_box_options(name, options)
+        check_box_options = options.except(:class, :label, :label_class, :error_message, :help,
+                                           :inline, :hide_label, :skip_label, :wrapper, :wrapper_class, :switch)
+        check_box_options[:class] = check_box_classes(name, options)
+        check_box_options[:aria] = { required: true } if options[:required]
```
lcreid, line 30: It took me a minute to figure out why we might want to make a check box required. I assume it's for cases where you won't accept the form unless the users checks something like: "I read and agree with the terms."

Would you mind adding a test to `test/bootstrap_checkbox_test.rb` that shows that the aria attribute is added correctly? I think the code is fine, but we need the tests to ensure that we don't introduce regressions in the future.

lib/bootstrap_form/form_group_builder.rb
```diff
@@ -77,11 +78,15 @@ def form_group_label_class(options)
...
-      options[:skip_required] ? false : :default
+    def form_group_required(options, method)
+      if options[:skip_required]
+        warn "`:skip_required` is deprecated, use `:required: false` instead"
+        false
+      elsif options.key?(:required)
+        options[:required]
+      else
+        required_attribute?(object, method)
+      end
```
lcreid, line 89: I like what you did here. I wonder how it interacts with https://github.com/bootstrap-ruby/bootstrap_form/blob/fe66580854058e243251ab6eabdbc42386ff68a5/lib/bootstrap_form/components/labels.rb#L28-L33. Is some of the code there now redundant?

lcreid (commented):
Thanks very much for the contribution. I really appreciate anything anyone can contribute for accessibility. 

It looks good so far. I think we'll be able to merge this after you address the comments.

lib/bootstrap_form/form_group_builder.rb
CharlieWinkwaves: I think you are right i changed it to only check for the options[:required] only one spec failed and that spec should not had have the required class anyway i think.

lib/bootstrap_form/inputs/check_box.rb
CharlieWinkwaves: You are correct i added a test for the checkbox, and also added the required class to the checkbox label

lib/bootstrap_form/components/labels.rb
```diff
@@ -25,13 +25,7 @@ def generate_label(id, name, options, custom_label_col, group_layout)
...
         classes = ["form-label", options[:class], label_layout_classes(custom_label_col, group_layout)]
 
-        case options.delete(:required)
-        when true
-          classes << "required"
-        when nil, :default
-          classes << "required" if required_attribute?(object, name)
-        end
-
+        classes << "required" if options.delete(:required)
```
lcreid, line 28: Just looking at this change, it looks like a change to the behaviour of the software. Why don't we care about the `nil, :default` case anymore?

lcreid (commented):
Thanks for adding the test. I have one more question. :-)

test/bootstrap_checkbox_test.rb
```diff
@@ -622,4 +622,15 @@ class BootstrapCheckboxTest < ActionView::TestCase
...
+  test "a required checkbox" do
+    expected = <<~HTML
+      <div class="form-check mb-3">
+        <input #{autocomplete_attr} name="user[terms]" type="hidden" value="0" />
+        <input aria-required="true" class="form-check-input" id="user_terms" name="user[terms]" required="required" type="checkbox" value="1"/>
+        <label class="form-check-label required" for="user_terms">I agree to the terms</label>
+      </div>
+    HTML
+    assert_equivalent_xml expected, @builder.check_box(:terms, label: "I agree to the terms", required: true)
+  end
```
lcreid, line 635: This is a good test. Could you also add a test where the attribute is required by a validation, rather than by giving the `required` argument to the `check_box` helper?

test/bootstrap_form_group_test.rb
```diff
@@ -316,7 +318,7 @@ class BootstrapFormGroupTest < ActionView::TestCase
 
     expected = <<~HTML
       <div class="mb-3 row">
-        <label class="form-label col-form-label col-sm-2 required" for="user_email">Custom Control</label>
+        <label class="form-label col-form-label col-sm-2" for="user_email">Custom Control</label>
```
lcreid, line 321: I believe this field is required. E-mail is required by the underlying model, so we want to make it a required field on the form, so this class would remain here, and you'd add the `aria-required` and `required` attributes to the `input`.

lcreid (commented):
Thanks for your patience while waiting for my review. Let me know if my comments don't make sense.

I also got to thinking, if we're adding "required" to check boxes, is there a similar concept for radio button? What do you think? If you think it makes sense, let me know. If it's easy, maybe you can do it in this PR? If you think it merits its own PR, please create an issue so we can make sure it gets done.

Thanks again for your efforts to improve `bootstrap_form`!!

test/bootstrap_form_group_test.rb
CharlieWinkwaves: You are right this was a bug in my code  I think I have fixed it

CharlieWinkwaves:
> 

Yes maybe good to do it as well for radio buttons. I think i created a better solution to make sure the same check is done everywhere.

lcreid (approved):
Looks good! Thanks for your patience while I reviewed this, and thanks again for all the work you've done to make this a better gem.

Issue #477, opened 2018-07-15
Add block to `collection_check_boxes` and `collection_radio_buttons`.


lcreid:
The Rails `collection_check_boxes` and `collection_radio_button` methods [can take a block](http://edgeapi.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-collection_check_boxes). If a block is provided, the helper yields a special builder to the block. The idea is that the programmer can format the check box or radio button more precisely in the block, if the standard format produced by the Rails helpers doesn't give the desired results.

Our helpers do *not* take a block. We can speculate that since our helpers format the check boxes and radio buttons in a consistent and pleasing way, there's less need for the block. There may also have been a technical reason that made yielding to a block difficult. I haven't investigated enough to have an idea of that.

However, there still may arise situations in which it would be convenient to have the ability to output the controls in a block. On example would be when you want to show all the check boxes or radio buttons, but some of them need to be disabled. Another example might be to highlight certain "dangerous" check boxes or radio buttons with the "danger" colour.

Another argument in favour of providing this is simply to make our helpers more closely mimic the behaviour of the Rails helpers. I bet there are a number of cases of people wasting an hour or two trying to get a block to work with our helpers before they realized that our helper doesn't work that way.

One alternative is for programmers to code the loop themselves. This is probably not a large burden, although they would have to remember to add the `div.form-group` and the error handling in order to be 100 percent compatible with our helpers.

Would we consider a pull request for this functionality if someone submitted one?

lcreid:
Here's a specific example I'm having trouble implementing: I have a `has_many :team_members, through:` on the `Subject` model. I want to provide a simple user interface with a check boxes for each person on staff. Checking the box makes them a member of the team. However, I want to prevent the user from removing themself from the team, so I don't want to show the check box for the current user.

In ordinary Rails I could do this:

```
<%= form_for @subject do |f| %>
  <%= f.collection_check_boxes(:team_member_ids, Person.all, :id, :name) do |builder| %>
    <% unless current_user == builder.object %>
      <%= builder.check_box %><%= builder.name %>
    <% end %>
  <p><%= f.submit %></p>
<% end %>
```

I can't seem to find a similarly elegant way to do this with `bootstrap_form`'s `collection_check_box`. @mattbrictson or anyone else, do you have any ideas for this? Am I missing something?

JoseFarias:
[...]

PR #665, opened 2022-11-01
allow for data- attributes to pass through for radio button collections

johrstrom:
Hi, I need the capability to add `data-` HTML attributes to collections of radio buttons. If found that this works, it may or may not be implemented the way you'd like - so I'm happy to fix/change whatever you ask (this PR is editable by maintainers). 

Just a little more information here, you can see from the tests I'm adding `data-city` attributes to each input tag.
```html
<input class="form-check-input" data-city="east" id="user_misc_1" name="user[misc][]" type="checkbox" value="1" />
```

We pass in `Array`s using `:first` and `:second`  for value and label functions respectively. One element, the `Hash`, is all the extra HTML `data-` attributes we'd like to add.
```ruby
['1', 'Foo', {'data-city': 'east'} ],
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,6 +739,28 @@ Collection methods accept these options:
 * `:help`: Add a help span to the `form_group`
 * Other options will be forwarded to the `radio_button`/`check_box` method
 
+
+To add `data-` attributes to a collection of radio buttons, map your models to an array and add a hash:
+
+```erb
+<%# Use the :first and :second elements of the array to be the value and label repsectively %>
+<%- choices = @collection.map { |addr| [ addr.id, addr.street, { 'data-zip-code':  addr.zip_code } ] } -%>
+
+<%= form.collection_radio_buttons :misc, choices, :first, :second %>
+```
+
+This generates:
+```html
+<div class="form-check">
+  <input data-zip-code="47805" class="form-check-input" type="radio" value="1" name="user[misc]" id="user_misc_1">
+  <label class="form-check-label" for="user_misc_1">Foo</label>
+</div>
+<div class="form-check">
+  <input data-zip-code="89807" class="form-check-input" type="radio" value="2" name="user[misc]" id="user_misc_2">
+  <label class="form-check-label" for="user_misc_2">Bar</label>
+</div>
+```
+
 ## Range Controls
 
 You can create a range control like this:
--- a/lib/bootstrap_form/inputs/inputs_collection.rb
+++ b/lib/bootstrap_form/inputs/inputs_collection.rb
@@ -29,7 +29,7 @@ module BootstrapForm
           input_options[:checked] = form_group_collection_input_checked?(checked, obj, input_value)
         end
 
-        # add things like 'data-' directives to the HTML
+        # add things like 'data-' attributes to the HTML
         obj.each do |inner_obj|
           input_options.merge!(inner_obj) if inner_obj.is_a?(Hash)
         end if obj.respond_to?(:each)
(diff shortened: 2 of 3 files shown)


lcreid (commented):
Thanks for the PR! It's always great to hear of new ideas that people have to make this gem better. 

Would you mind adding a change to the `README.md` file that would tell a programmer how they would use your feature to achieve something they want to do? I find that helps me understand your use case. Thanks in advance.

johrstrom:
> Would you mind adding a change to the README.md file

Of course! I added it to the `collections` section as this only applies to collections (IDK if folks can already do this with a single `radio_button`). I didn't add an image because visually it's the same as before, but I'm happy to if you want one.

johrstrom:
Is there something wrong with the example and/or would you like other updates?

lcreid:
Thank you for your patience. I really had to do some updating of personal projects/servers the last little while, and that sucked up all the time I had available for non-paid work. :-(

The reason this is complicated is because, while I like your solution, it may bring to light a bigger [issue](#477 ). I'm guessing that the way to add attributes to the individual check boxes (and radio buttons) in a more general way, if you were using the [underlying Rails helper](https://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-collection_check_boxes), would be to provide a block.  and then you have much more control over the rendering of the individual check boxes, including adding whatever data attributes you needed.

Unfortunately, I already have a big refactoring that I think I need to do to solve another issue, so I can't promise to work on #477 in the next few weeks. So here's what I'd like to propose (and they're not mutually exclusive):

- Merge your PR, but mark it as a experimental feature that is _very likely_ to change or not be supported in the future.
- Explore ways to fix #477 and how that would solve your issue, among others.

I'd be really happy to see what you think about the second one. We would have to have a discussion about it (i.e. don't run off and implement a whole PR just yet). But I understand if you don't have time to take it on. At least the first one point would allow you to move on.

What do you think, @johrstrom ?

johrstrom:
Thanks for the response. I'm happy to look into providing blocks. I work in OSS as a part of my job (this if for that job) so I can make the time.

You're the maintainer, and as a fellow maintainer of OSS, I get that you may not want to take on additional features when you know/suspect they'll be replaced. So it's totally fine by me to close this PR and only look into providing blocks.  What's more is is that my project is on bootstrap 4, so even if this were merged, I won't update and use it for quite some time. So there's no urgency.

> We would have to have a discussion about it (i.e. don't run off and implement a whole PR just yet). 

Happy to continue the discussion on #477

johrstrom:
Looking into #477 and the Rails library, I have to set the expectation here that I doubt I'll get very far in implementing it. And if I do it'd be quite some time. It's not really an issue of time (well it is kinda) it's more of the complexity of the change and my familiarity with this library.

lcreid:
I hear you. I think what has happened is that the structure of the code meshed well with Bootstrap 3, and worked okay with Bootstrap 4, now we're at Bootstrap 5, and we've been adding features all that time, and the code is getting hard to change. There's another issue that I simply can't fix because I can't find an incremental way to get the code there. And I just don't have the time for a rewrite.

If I manage to do some work, do you mind if I mention you for reviews or with questions? I find other eyes can be really helpful.

johrstrom:
> If I manage to do some work, do you mind if I mention you for reviews or with questions? I find other eyes can be really helpful.

Sure thing! Like I say I'm a maintainer of an OSS package, though through my job. In any case, I have incentive to help you because I need this library too, so I'm very happy to do so.

donv:
Hi @johrstrom and @lcreid !

What is the state of this PR now?

johrstrom:
> What is the state of this PR now?

I don't know. Reading through the comments it seems the TLDR is `it's OK, but didn't get merged because a larger refactor is preferred`.  I cannot initiate such a refactor, but would be happy to help if I can.

That said, if you close this PR my feelings won't be hurt. If you all want to close this in preference for a larger refactor, that's totally understandable.

donv:
The change is small and well tested and documented, so I merged it.

johrstrom:
Thanks!

PR #667, opened 2022-12-06
Update gemspec

siegy22:
The current gem packages a whole rails application (`demo/`) in the `.gem` file, I assume this was added by mistake 🤔


lcreid (approved):
Thanks for the PR! Perhaps at one time there was a thought that you should be able to develop for the gem just by downloading the gem, but that's certainly not the way we do things anymore. I think this PR is the right thing to do. Thanks again!

Issue #674, opened 2023-04-06
Tests May Not Be Good Enough


lcreid:
The current way we're testing may be too loose in checking the HTML. It looks like XML test we're using isn't strict enough about the order of elements. I discovered that I can make the check box tests pass, even thought the label is being put in front of the check box, instead of after it.

The first step is to reproduce the problem and confirm that it's a real problem.

PR #676, opened 2023-04-14
674 html tests

lcreid:
In working on a new implementation, it began to look like tests were passing when they shouldn't have passed. I made a branch off `main` and confirmed that we weren't testing the order of elements, which is significant in HTML.

Once that was fixed, it turned out that our code for handling some of the variations was also passing tests that it shouldn't have been passing. A couple of more changes fixed that.

Finally, three tests were failing because the expected strings were incorrect, but the other issues mentioned above were hiding that the tests should have failed.

In the process of fixing, we changed to use Nokogiri's HTML instead of XML. And to make it clear, the assertion helper was changed in all the test files.

Fixes: #674

### DIFF
--- a/test/bootstrap_checkbox_test.rb
+++ b/test/bootstrap_checkbox_test.rb
@@ -15,7 +15,7 @@ class BootstrapCheckboxTest < ActionView::TestCase
         </label>
       </div>
     HTML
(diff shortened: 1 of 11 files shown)


donv (approved):
Nice!

PR #679, opened 2023-05-19
Update ruby requirement to >=3.0

donv:
Ruby 2.7 is EOL.


lcreid (approved):
Yes, it's definitely time. Thanks for this!

PR #682, opened 2023-05-20
Set correct `form-range` class for range_field

### LINKED ISSUE #677: Range control has wrong class
`<%= f.range_field :excellence %>` generates:

```html
<div class="mb-3">
  <label class="form-label" for="user_excellence">Excellence</label>
  <input class="form-control" id="user_excellence" name="user[excellence]" type="range">
</div>
```

According to the [Bootstrap 5 docs](https://getbootstrap.com/docs/5.3/forms/range/), the input should have the class `form-range`, not `form-control`.

donv:
Fixes #677


lcreid (approved):
Thanks for this fix!

PR #684, opened 2023-05-23
Allow ActiveModel object as form objects.

### LINKED ISSUE #683: Need to include ActiveRecord::Reflection for custom form class with 5.2.0
With release 5.2.0 my app's system tests run into this error:

```
     ActionView::Template::Error:
       undefined method `reflections' for App::UserForm:Class
```

I believe this is after this change: https://github.com/bootstrap-ruby/bootstrap_form/pull/654

`App::UserForm` is a custom form class I made that loads and saves data from and to a few different ActiveRecord classes. To make this work with validations, so far I've only needed to `include ActiveModel::Model`:

```
module App
  class UserForm
    include ActiveModel::Model

    validates :street, :city, :country, presence: true
...
```

If I also `include ActiveRecord::Reflection` in my custom form class, the error is gone. What I'm wondering is if this is the right approach, or that maybe bootstrap_form should not rely on `reflections` being available? Or are there other thoughts for fixing this? And if this is the right approach, should this be documented as a change/solution on the side of bootstrap_form?

donv:
Fixes #683


thimo:
Awesome, thanks!

PR #685, opened 2023-05-23
Don't set required and aria-required for attributes with conditional presence validators

AndersGM:
After the nice addition in: https://github.com/bootstrap-ruby/bootstrap_form/pull/664 I've run into issues with attributes that have conditional presence validators (:if and :unless). This PR fixes that, so the required HTML-attributes are only added when we are sure the attributes are always required.


AndersGM:
@donv Would it be possible to make a new release with this fix? :-)

donv:
Done.

PR #686, opened 2023-06-18
Improve checkbox layout

donv:
* Indent checkbox into second column for horizontal layout
* Allow overriding the label class, not just add more classes.
* Align label with the top of the first check box in a collection in horizontal layout
* Vertically align check boxes and labels in inline layout.

Since there are breaking changes, overriding the label class and indenting the check box into the second column for horizontal layout, I would like to bump the version number to 5.3 also to indicate that we now support Bootstrap 5.3.


demo/doc/screenshots/bootstrap/index/00_horizontal_form.png
lcreid, line 1: If I did want the "Agree to Terms" check box to be flush left as it is in the old version, how would I do that?

demo/doc/screenshots/bootstrap/index/02_inline_form.png
lcreid, line 1: I really like that you improved the alignment in this and the following examples.

README.md
```diff
@@ -1096,9 +1098,7 @@ In the example below, the checkbox and submit button have been wrapped in a
 <%= bootstrap_form_for(@user, layout: :horizontal, label_col: "col-sm-2", control_col: "col-sm-10") do |f| %>
   <%= f.email_field :email %>
   <%= f.password_field :password %>
-  <%= f.form_group do %>
-    <%= f.check_box :remember_me %>
-  <% end %>
+  <%= f.check_box :remember_me %>
```
lcreid, line 1101: The text above says that the check box is wrapped in a form group for alignment, when this is apparently no longer necessary. I think the new code is fine but the text should be fixed.

test/bootstrap_form_test.rb
```diff
@@ -179,13 +181,17 @@ class BootstrapFormTest < ActionView::TestCase
...
+            <div class="form-check">
+              <input #{autocomplete_attr} name="user[terms]" type="hidden" value="0" />
+              <input class="form-check-input" id="user_terms" name="user[terms]" type="checkbox" value="1" />
+              <label class="form-check-label" for="user_terms">I agree to the terms</label>
+            </div>
+          </div>
         </div>
         <div class="mb-3 row">
-          <label class="form-label col-form-label col-sm-2" for="user_misc">Misc</label>
+          <label class="col-form-label pt-0 col-form-label col-sm-2" for="user_misc">Misc</label>
```
lcreid, line 194: Any idea why we expect `col-form-label` twice?

test/bootstrap_form_test.rb
```diff
@@ -228,13 +234,17 @@ class BootstrapFormTest < ActionView::TestCase
...
+          <div class="col-sm-10 offset-sm-2">
+            <div class="form-check">
+              <input #{autocomplete_attr} name="user[terms]" type="hidden" value="0" />
+              <input class="form-check-input" id="user_terms" name="user[terms]" type="checkbox" value="1" />
+              <label class="form-check-label" for="user_terms">I agree to the terms</label>
+            </div>
+          </div>
         </div>
         <div class="mb-3">
-          <label class="form-label" for="user_misc">Misc</label>
```
lcreid, line 237: I wonder why sometimes it wasn't `col-form-label` before?

lcreid (commented):
I share your concern about these being breaking changes, but since we're really fixing stuff that was broken before (mis-aligned fields and labels, for example), I think making the change is reasonable.

I'd like to see what you think about my comments and questions before I approve, but overall I really like what I'm seeing.

demo/doc/screenshots/bootstrap/index/00_horizontal_form.png
UweKubosch: Just added two ways to do it:  setting `layout: :default` or `label_col: ""`.

README.md
UweKubosch: Fixed the text now.

test/bootstrap_form_test.rb
UweKubosch: Not sure.

lcreid:
The edge test failures are #691

test/bootstrap_form_test.rb
lcreid: Looking more closely, I see that before we said, `form-label col-form-label`. [Current Bootstrap documentation](https://getbootstrap.com/docs/5.3/forms/layout/#horizontal-form-label-sizing) indicates we only need `col-form-label`, and only once (I could be wrong). Do you think you could clean that up so we only get `col-form-label` and only once?

UweKubosch: I will take a look 👍

donv: Reworked the code to only use the new label classes.

test/bootstrap_form_test.rb
donv: I found a place where the field layout option was not forwarded to the layout label classes method.  Fixed it.

donv:
@lcreid Good to merge?

donv:
I'll go ahead and merge this, then.

Laykou:
I've noticed this makes it difficult to add `help` 

Previously I could do:

```
= f.form_group :is_requestable do
    = f.check_box :is_requestable
    .form-text
      = t('activerecord.attributes.group.is_requestable_help') 
```

Now I need to remove all `f.form_group` since the offcet is now done twice - wrapped also as part of `f.check_box`
image


This now changed and `form_group` is removed (however it is still mentioned in the readme here: https://github.com/bootstrap-ruby/bootstrap_form/tree/main?tab=readme-ov-file#checkboxes-and-radios

Now it's enough to set just:

```
= f.check_box :is_requestable
```

However there is no way how to add additional text like

```
= f.check_box :is_requestable, help: 'Some explanatino'
```

PR #690, opened 2023-08-08
Fixed `Calling << to an ActiveModel::Errors` deprecation warning

### LINKED ISSUE #689: DEPRECATION WARNING: Calling `<<` to an ActiveModel::Errors message array in order to add an error is deprecated
When using `bootstrap_form` > `v5.1.0` with rails `v6.1.7.4` i receive the following deprecation warning:

`DEPRECATION WARNING: Calling `<<` to an ActiveModel::Errors message array in order to add an error is deprecated. Please call `ActiveModel::Errors#add` instead.`

I believe its coming from this line:

https://github.com/bootstrap-ruby/bootstrap_form/blob/main/lib/bootstrap_form/components/validation.rb#L74

dannytip:
Fixes #689

### DIFF
--- a/lib/bootstrap_form/components/validation.rb
+++ b/lib/bootstrap_form/components/validation.rb
@@ -65,6 +65,7 @@ module BootstrapForm
         content_tag(help_tag, help_text, class: help_klass)
       end
 
+      # rubocop:disable Metrics/AbcSize
       def get_error_messages(name)
         object.class.try(:reflections)&.each do |association_name, a|
           next unless a.is_a?(ActiveRecord::Reflection::BelongsToReflection)
@@ -77,6 +78,7 @@ module BootstrapForm
 
         object.errors[name].join(", ")
       end
+      # rubocop:enable Metrics/AbcSize
     end
   end
 end
--- a/lib/bootstrap_form/components/validation.rb
+++ b/lib/bootstrap_form/components/validation.rb
@@ -66,14 +66,16 @@ module BootstrapForm
       end
 
       def get_error_messages(name)
-        messages = object.errors[name]
         object.class.try(:reflections)&.each do |association_name, a|
           next unless a.is_a?(ActiveRecord::Reflection::BelongsToReflection)
           next unless a.foreign_key == name.to_s
 
-          messages << object.errors[association_name]
+          object.errors[association_name].each do |error|
+            object.errors.add(name, error)
+          end
         end
-        messages.join(", ")
+
+        object.errors[name].join(", ")
       end
     end
   end
--- a/demo/app/controllers/addresses_controller.rb
+++ b/demo/app/controllers/addresses_controller.rb
@@ -1,9 +1,0 @@
-class AddressesController < ApplicationController
-  def create
-    redirect_to root_path
-  end
-
-  def update
-    redirect_to root_path
-  end
-end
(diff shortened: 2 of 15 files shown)


dannytip:
This is currently failing linting due to:

`Metrics/AbcSize: Assignment Branch Condition size for get_error_messages is too high. [<3, 18, 6> 19.21/18]`

Any suggestion on how to resolve that @lcreid ?

lcreid (commented):
Thanks for submitting a PR. We should have a test that shows that this is fixed. I believe our test suite wasn't triggering this code before, which is why we never noticed the error. Do you have a test that can reproduce the error (before your fix) that we could add to our tests?

lcreid:
> This is currently failing linting due to:
> 
> `Metrics/AbcSize: Assignment Branch Condition size for get_error_messages is too high. [<3, 18, 6> 19.21/18]`
> 
> Any suggestion on how to resolve that @lcreid ?

I wonder why that's come up for you? Perhaps that's why the original code used a temporary variable? For now, put `# rubocop:disable Metrics/AbcSize` at the start of the method, and `# rubocop:enable Metrics/AbcSize` at the end of the method.

dannytip:
@lcreid ive added a test case which triggers the warning when you change `REQUIRED_RAILS_VERSION` to `v6.1.7.4` strangely you don't get the warning with rails `v7.0.6`.

lib/bootstrap_form/components/validation.rb
```diff
@@ -65,16 +65,20 @@ def generate_error(name)
...
+      # rubocop:disable Metrics/AbcSize
       def get_error_messages(name)
-        messages = object.errors[name]
         object.class.try(:reflections)&.each do |association_name, a|
           next unless a.is_a?(ActiveRecord::Reflection::BelongsToReflection)
           next unless a.foreign_key == name.to_s
 
-          messages << object.errors[association_name]
+          object.errors[association_name].each do |error|
+            object.errors.add(name, error)
```
lcreid, line 75: I was intrigued by your comment that it didn't say anything at version 7. You would think it should have errored, if the deprecation warning was in 6.1. You're right. `<<` seems to still work, and without a deprecation message. Very strange.

I was unable to get it to show me the deprecation message, even with version 6.1. (I tweaked the code to run your test case, but with the `<<` operator instead of `#add`.) Also strange. At any time, did you see the deprecation warning when running the test you added?

lcreid (commented):
This looks great. I'm interested in seeing your response to my comments before approving, but we should be good to go after that.

lcreid:
Oops. Looks like the test fails on edge.

lib/bootstrap_form/components/validation.rb
dannytip: If i revert my changes to `get_error_messages` but leave the new testcase in place i get the following:

Screenshot 2023-08-10 at 09 28 03

dannytip:
> Oops. Looks like the test fails on edge.

It looks like it doesn't pick up that the relationship is required in rails 7.1.  Possibly something in `required_association?`

lcreid (approved):
Thanks for removing the unnecessary code. 

We can merge this with failing tests in edge. I've created [an issue](#691) for that failure. The issue has a lead about what may be the problem.

PR #694, opened 2023-08-30
Rename layout `:default` to `:vertical`

### LINKED ISSUE #687: Rename the default layout to "vertical" ?
As stated in the [Bootstrap docs](https://getbootstrap.com/docs/5.3/forms/layout/#forms):

"forms will by default stack vertically"

The current name, "default", does not actually hold any meaning.  Also, the default can be changed to "inline" or "horizontal".

I propose we change the name of the default layout to "vertical".

What do you think?

donv:
Fixes #687


donv:
I'd like to move towards a release, so I'll merge this change.  @lcreid hope this is OK.

PR #695, opened 2023-09-06
Remove aria-required attribute from semantic element `input`

jdufresne:
Per the MDN docs, the aria-required attribute is only necessary on non-semantic elements. The `input` element is a semantic element and so aria-required is not necessary -- just `required` is sufficient and achieves the same result. An example of a non-semantic element is `div` which is not used by the project.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required

> When a semantic HTML <input>, <select>, or <textarea> must have a
> value, it should have the required attribute applied to it.
>
> ...
>
> When form controls are created using non-semantic elements, such as a
> <div> with a role of checkbox, the aria-required attribute should be
> included, with a value of true, to indicate to assistive technologies
> that user input is required on the element for the form to be
> submittable.

By removing the unnecessary attribute, there is less noise in the rendered HTML and less total HTML.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ This generates the following HTML:
 <form accept-charset="UTF-8" action="/users" class="new_user" id="new_user" method="post">
   <div class="mb-3">
     <label class="form-label required" for="user_email">Email</label>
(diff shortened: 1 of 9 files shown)


donv:
This looks OK.

PR #699, opened 2023-09-08
Add `bootstrap_fields_for` and `bootstrap_fields` form helpers.

### LINKED ISSUE #283: Non-nested fields_for
Currently you can use fields_for to create a nested form like this:

<%= bootstrap_form_for @person do |person_form| %>
  <%= person_form.fields_for @person.address do |address_fields| %>
      <%= address_fields.text_field :street %>
  <% end %>
  <%= person_form.submit %>
<% end %>

This way the input :street wil get the name person[address][street]. I would like that it get the name address[street] so not nested inside person

Using the Rails form helper you would do this by changing  "person_form.fields_for" to simply "fields_for" (which works like form_for but without the creation of enclosing form tags)

However bootstrap_fields_for or something similar does not seem to exist. 

Is there maybe an option you can pass to bootstrap_form_for so that it will not output the form tags?

donv:
Fixes #283


README.md
```diff
@@ -164,6 +164,37 @@ in `form_with`.
 
 `form_with` has some important differences compared to `form_for` and `form_tag`, and these differences apply to `bootstrap_form_with`. A good summary of the differences can be found at: https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a, or in the [Rails documentation](api.rubyonrails.org).
 
+### bootstrap_fields_for
+
+Adding fields for a different object without nesting can be achieved using the `bootstrap_fields_for` helper in the same way it is done in rails.
```
lcreid, line 169:
```suggestion
Adding fields for a different object without nesting can be achieved using the `bootstrap_fields_for` helper in the same way it is done in Rails.
```

lcreid (commented):
Thanks for all the work you've been doing lately!

We should have test cases for the new method -- I expect a few tests, since this method has optional arguments, and our long list of options. I not proposing we completely re-test `BootstrapForm::FormBuilder`, but maybe consider some of the key use cases for this method and where they might interact with how our form builder works.

I'm curious about the `:builder` option. I quickly scanned the Rails docs but didn't see an explicit mention. Did I just miss it?

donv:
> We should have test cases for the new method -- I expect a few tests, since this method has optional arguments, and our long list of options. I not proposing we completely re-test `BootstrapForm::FormBuilder`, but maybe consider some of the key use cases for this method and where they might interact with how our form builder works.

I have added a couple of tests to see that it works across the test matrix.  The new method actually has no options of itself, it only forwards arguments to the upstream `fields_for` method, so I don't see what variants are useful to test.  I also added the `bootstrap_fields` method since it is trivial to implement.

> I'm curious about the `:builder` option. I quickly scanned the Rails docs but didn't see an explicit mention. Did I just miss it?

The concrete option is not mentioned in the method documentation comment.  The builder is barely mentioned, but is clearly there in the code:

https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/form_helper.rb#L1614C11-L1614C11

donv:
> I'm curious about the `:builder` option. I quickly scanned the Rails docs but didn't see an explicit mention. Did I just miss it?

I found an example here:  https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/form_helper.rb#L388

lcreid:
Thanks for digging up some details about the `builder` argument. Maybe the assumption is that one understands that all those helpers are similar to the `form_for` helper that has the `builder` argument.

Also, I appreciate you thinking through the testing. I suspected it might be hard to determine what to test, but sometimes I think it's useful to ask people to take one more run at the problem before putting it to bed. Thanks for doing the thinking and researching.

donv:
> Thanks for digging up some details about the `builder` argument. Maybe the assumption is that one understands that all those helpers are similar to the `form_for` helper that has the `builder` argument.
> 
> Also, I appreciate you thinking through the testing. I suspected it might be hard to determine what to test, but sometimes I think it's useful to ask people to take one more run at the problem before putting it to bed. Thanks for doing the thinking and researching.

My pleasure!  😄

PR #704, opened 2023-09-22
Fix joins not to make strings unsafe

lcreid:
The basic `String#join` makes its output unsafe in all cases. This might have been causing unwanted and problematic escaping in some cases. This PR removes the use of `String#join`. It _should_ still escape unsafe strings passed in from the calling code, while not escaping anything that is already safe. However, that's something reviewers should put some thoughts into, and comment on.

This PR may cause some text to be escaped that wasn't previously escaped, but those cases would have been potentially security vulnerabilities, so I think it's justifiable in closing that possible loophole. The fix would be for the calling code to pass its string as `string.html_safe` if the code is confident the string is safe (i.e. it doesn't come from the user or the database or something external). If the string wasn't safe, then it _should_ get escaped.

### DIFF
--- a/lib/bootstrap_form/components/validation.rb
+++ b/lib/bootstrap_form/components/validation.rb
@@ -81,7 +81,8 @@ module BootstrapForm
           end
         end
 
-        object.errors[name].join(", ")
+        safe_join(object.errors[name], ", ")
+        # object.errors[name].join(", ")
       end
       # rubocop:enable Metrics/AbcSize
     end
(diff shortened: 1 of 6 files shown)


lib/bootstrap_form/components/validation.rb
```diff
@@ -81,7 +81,8 @@ def get_error_messages(name)
           end
         end
 
-        object.errors[name].join(", ")
+        safe_join(object.errors[name], ", ")
+        # object.errors[name].join(", ")
```
UweKubosch, line 85: Remove  commented code.

UweKubosch (approved):
Looks good to me 😄

PR #706, opened 2023-09-29
rescue instead if actiontext could not be found

BenMorganMY:


lib/bootstrap_form.rb
```diff
@@ -1,9 +1,11 @@
-# NOTE: The rich_text_area and rich_text_area_tag helpers are defined in a file with a different
-# name and not in the usual autoload-reachable way.
+# NOTE: The rich_text_area and rich_text_area_tag helpers are defined in a file
+# with a different name and not in the usual autoload-reachable way.
 # The following line is definitely need to make `bootstrap_form` work.
-if Rails::VERSION::STRING > "6"
+begin
   require "#{Gem::Specification.find_by_name('actiontext').gem_dir}/app/helpers/action_text/tag_helper"
+rescue Gem::MissingSpecError
```
lcreid, line 6: I can see why this is a better solution. I wonder if it would be worthwhile writing an error message in the `rescue` case. It might be better than silently swallowing the error. (Admittedly this could also be said of the current code.) What do you think?

lcreid (commented):
Thanks for this suggestion! It makes a lot of sense to me. We really appreciate everyone's efforts to make the gem better. I had one question before we merge this PR.

lib/bootstrap_form.rb
BenMorganMY: That would mean that any application that is loaded without actiontext would always get an error message. It might be more useful to put error handling when the user attempts to use the action text tag helper.

lcreid: Of course. Good point. And when they try to use the action text helper, they'll get a obvious (by Rails standards, at least) error message. Thanks for walking me through the thinking!

lcreid:
You can ignore the failures against edge Rails, but please fix up the lint issue by ignoring the RuboCop warning.

lcreid (approved):
Thanks again for this PR!

PR #708, opened 2023-10-02
Rails 7.1

lcreid:
Add Rails 7.1 to the matrix. Also add `ruby-head` to the matrix.

This was remarkably painful to finish. A [small change in the behaviour](https://blog.saeloun.com/2023/02/02/rails-allow-db-prepare-to-load-schema-if-database-exists/) of `rails db:prepare` broke our tests, and that sent me off on a adventure through a couple of other PRs that were supposed to help me get the tests running for Rails 7.1, but ended up being a lot of work in themselves. Sigh.

Most of the changes in this PR are files generated by the Rails upgrade of the demo app.

One of the things I tried was simply ignoring the new table, hoping that that would fix (or at least delay) the need to deal with the change to `db:prepare`, but the model, and therefore the table, is referenced somewhere deep in the Rails initialization code.

Also of note: I don't think anything about Rails 7.1 actually broke the gem. The problems I had or caused myself were because of the way we test the gem.

I'll do a release to 5.4.0 after merging this PR.


CONTRIBUTING.md
```diff
@@ -85,13 +85,20 @@ Docker is _not_ required to work on this gem.
 
 The `docker-compose` approach should link to enough of your networking configuration that you can release the gem.
 However, you have to do some of the configuration yourself, because it's dependent on your host operating system.
+
```
donv, line 88: I feel the Docker documentation has grown so large it should be extracted into its own document (DOCKER.md) and then be referred from the CONTRIBUTING.md document.

demo/Gemfile
```diff
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 gem "bootstrap_form", path: ".."
 
-gem "rails", "~> 7.0.0"
+gem "rails", "~> 7.1.0"
```
donv, line 5: Might be a good idea to go directly to 7.1.1 since it contains bugfixes.

demo/Gemfile.lock
```diff
@@ -241,12 +262,12 @@ DEPENDENCIES
...
+  rails (~> 7.1.0)
   selenium-webdriver
   sprockets-rails
   sqlite3 (~> 1.4)
   tzinfo-data
   web-console
 
 BUNDLED WITH
-   2.4.13
+   2.4.1
```
donv, line 273: I think it is a good idea to keep up with newer versions of bundler using `bundle update --bundler`.  At least not downgrade bundler without a reason.

demo/config/application.rb
```diff
@@ -1,26 +1,28 @@
 require_relative "boot"
 
 require "rails/all"
-# require "sprockets/railtie"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
-require "bootstrap_form"
+require "bootstrap_form" # FIXME: Why do we require this here?
```
donv, line 8: Have you tried without this line?

donv (approved):
Looks good to me.

demo/config/application.rb
lcreid: Argh. I was sure I had removed the FIXME. Anyway, yes, I have another branch where I'm doing some small fix-ups, and it seems to run fine without this `require`. Maybe I'll merge that branch before I release.

demo/Gemfile.lock
lcreid: Agreed. I was using an older Docker container. Sometimes it seems like a stable development platform can cause more problems than it solves. :frowning_face:

CONTRIBUTING.md
lcreid: Good idea. I'll do that in a separate PR.

PR #709, opened 2023-10-05
do not alias rich text area if it is not defined

BenMorganMY:
We experienced an error with eager loading in staging for bootstrap_form. This small if statement resolves that issue.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,9 +37,6 @@ Layout/LineLength:
 Layout/SpaceAroundEqualsInParameterDefault:
   EnforcedStyle: no_space
 
-Lint/SuppressedException:
-  Enabled: false
-
 Metrics/AbcSize:
   Max: 18
   Exclude:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,6 +37,9 @@ Layout/LineLength:
 Layout/SpaceAroundEqualsInParameterDefault:
   EnforcedStyle: no_space
 
+Lint/SuppressedException:
+  Enabled: false
+
 Metrics/AbcSize:
   Max: 18
   Exclude:
@@ -45,6 +48,7 @@ Metrics/AbcSize:
 
 Metrics/BlockLength:
   Exclude:
+    - "lib/bootstrap_form/inputs/base.rb"
     - "demo/config/**/*"
     - "demo/test/**/*"
     - "test/**/*"
--- a/lib/bootstrap_form/inputs/base.rb
+++ b/lib/bootstrap_form/inputs/base.rb
@@ -40,7 +40,7 @@ module BootstrapForm
         def bootstrap_alias(field_name)
           alias_method "#{field_name}_without_bootstrap".to_sym, field_name
           alias_method field_name, "#{field_name}_with_bootstrap".to_sym
-        rescue NameError # rubocop:disable Lint/SupressedException
+        rescue NameError # rubocop:disable Lint/SuppressedException
         end
       end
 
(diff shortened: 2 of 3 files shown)


lcreid (commented):
Thanks for this PR. I really appreciate the time you're dedicating to make `bootstrap_form` better. I'm no expert on how Rails loads, but I'm guessing that when the tests run, our code is being run before `rich_text_area` is defined.

I really think we should get this to work the way you want it, but also have all the tests passing. Feel free to take a swing at it, or just comment back that you want someone else to take a run at it.

BenMorganMY:
@lcreid I've switched from using the check to now just silencing the `NameError` that gets raised when there's a bad alias. I've opted to disable the lint for it and add docs explaining the decision on why it's there. I see this as the simplest way forward.

lib/bootstrap_form/inputs/base.rb
```diff
@@ -31,9 +31,16 @@ def bootstrap_select_group(field_name)
...
+        # Creates the methods *_without_bootstrap and *_with_bootstrap.
+        #
+        # If your application did not include the rails gem for one of the dsl
+        # methods, then a name error is raised and suppressed. This can happen
+        # if your application does not include the actiontext dependency due to
+        # `rich_text_area` not being defined.
         def bootstrap_alias(field_name)
           alias_method "#{field_name}_without_bootstrap".to_sym, field_name
           alias_method field_name, "#{field_name}_with_bootstrap".to_sym
+        rescue NameError
```
lcreid, line 43: I'd prefer to do the RuboCop exception specifically on the line.
```suggestion
        rescue NameError # rubocop:disable Lint/SupressedException
```

.rubocop.yml
```diff
@@ -37,6 +37,9 @@ Layout/LineLength:
 Layout/SpaceAroundEqualsInParameterDefault:
   EnforcedStyle: no_space
 
+Lint/SuppressedException:
+  Enabled: false
+
```
lcreid, line 42: See below.
```suggestion
```

.rubocop.yml
```diff
@@ -45,6 +48,7 @@ Metrics/AbcSize:
 
 Metrics/BlockLength:
   Exclude:
+    - "lib/bootstrap_form/inputs/base.rb"
```
lcreid, line 51: I guess we can leave this one in here, since it applies to the whole file anyway. But I should look at the rule and where it came from. I'm a fan of RuboCop, but not of these rules that put arbitrary limits on the size of things.

lcreid (commented):
I appreciate the effort you're putting into this. I think your approach is reasonable. I'm going to take a look at the problem as soon as I can and see if there's just something about the way we're loading that should be looked at. I suspect there is. This is not the first time we've had issues around loading the tests.

BenMorganMY:
@lcreid updated the rubocop lint suppression suggestion.

lib/bootstrap_form/inputs/base.rb
```diff
@@ -31,9 +31,16 @@ def bootstrap_select_group(field_name)
...
+        # Creates the methods *_without_bootstrap and *_with_bootstrap.
+        #
+        # If your application did not include the rails gem for one of the dsl
+        # methods, then a name error is raised and suppressed. This can happen
+        # if your application does not include the actiontext dependency due to
+        # `rich_text_area` not being defined.
         def bootstrap_alias(field_name)
           alias_method "#{field_name}_without_bootstrap".to_sym, field_name
           alias_method field_name, "#{field_name}_with_bootstrap".to_sym
+        rescue NameError # rubocop:disable Lint/SupressedException
```
lcreid, line 43:
```suggestion
        rescue NameError # rubocop:disable Lint/SuppressedException
```

lcreid (approved):
Thanks for this PR!

PR #710, opened 2023-10-21
Dockerize System Tests

lcreid:
@donv very nicely added system tests that generate screenshots for the `README`. This PR adds support to the optional Docker and Docker Compose files so that the system tests can be run in Docker.

The intent is to ensure that system tests can still be run without Docker, but that they will also run for those who prefer to develop in a Docker environment, because they want to isolate their development work, or for whatever other reason.

This PR introduces a new Docker Compose file: `docker-compose-system-test.yml`. This file uses a standard Selenium image for Chrome to run the browser. Part of the reason for a separate file is the current `docker-compose.yml` results in smaller containers, since it doesn't have the full browser. I also wanted to try things in a completely separate file, to avoid breaking the existing Docker workflow.

The approach in this PR is highly experimental and includes some rough edges:

- Actually running the tests is manual. There are a few steps to it, and while they may be familiar to people used to working with Docker, they're still inconvenient.
- As documented, files generated by the tests are owned by `root`. There may be a simple fix for this, but I haven't tried it yet.
- I haven't confirmed that system tests still work for non-Docker users.
- The screenshots generated by the Docker workflow are one pixel bigger on the x-axis than the ones stored in the repository.
- The locale needs to be set in the Selenium container to be consistent with the existing screenshots.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,14 @@
-.bundle/
+.npm/_update-notifier-last-checked
+.npm/_cacache/
+.npm/_logs/
+.npm/_npx
+
 .idea
 log/*.log
 pkg/
 demo/db/*.sqlite3
+demo/db/*.sqlite3-shm
+demo/db/*.sqlite3-wal
 demo/doc/screenshots/**/*.diff.png
 demo/log/*.log
 demo/tmp/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,7 +123,26 @@ services:
 
 You may have to change the `1000:1000` to the user and group IDs of your laptop. You may also have to change the `version` parameter to match the version of the `docker-compose.yml` file.
 
-Adapting the above `docker-compose.override.yml` for MacOS should be relatively straight-forward. Windows users, I'm afraid you're on your own.
+Adapting the above `docker-compose.override.yml` for MacOS should be relatively straight-forward. Windows users, I'm afraid you're on your own. If you figure this out, a PR documenting how to do it would be most welcome.
+
+The above doesn't allow you to run the system tests. To keep the image small, it doesn't include Chrome or any other browser.
+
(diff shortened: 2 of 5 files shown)


lcreid:
@donv would you mind pulling this branch and testing if it still runs system tests however you were running them before?

lcreid:
I think it may be a dream to make the system tests work on all possible platforms people could be using, since none of them have exactly the same fonts. At least, I haven't seen a good solution to this problem. Let me know if you've seen screenshot tests that work across platforms.

lcreid:
I should have mentioned that I have been using Rails 7.0.8 while working on this PR.

lcreid:
@donv Did you generate the screenshots for the `README` on a Mac, or Windows, or Linux (distro?)?

donv:
> @donv Did you generate the screenshots for the `README` on a Mac, or Windows, or Linux (distro?)? The reason I ask is that I'm trying to find a font on Linux that will be close (or the same) metrics-wise to your screenshots, so we can at least visually scan the screenshots quickly for significant differences.

I generated the screenshots on MAC.

I have recently, in another project, implemented a scheme for generating the screenshots in GitHub actions and updating them using pull requests from a separate branch for screenshots.  This seems to work very well, and I think that should be the way forward.

donv (approved):
The changes look good to me, and all tests are green in my workspace using your fork+branch.

lcreid:
> I have recently, in another project, implemented a scheme for generating the screenshots in GitHub actions and updating them using pull requests from a separate branch for screenshots. This seems to work very well, and I think that should be the way forward.

Brilliant! I've created an [issue](https://github.com/bootstrap-ruby/bootstrap_form/issues/711) for this, and assigned it to myself. But I'll probably un-assign myself in a day or two, when I don't find time to work on it. Feel free to assign yourself.

Issue #711, opened 2023-10-28
Generate System Test Baseline via GitHub Actions


lcreid:
For a gem like this, it's really useful to be able to actually test the rendered output of Rails. @donv did some awesome work to develop processes that generate screenshots directly from the `README.md`, and compare them to expected screenshots in the repo. 

Unfortunately, the expected screenshots highly depend on the platform (e.g. browser and fonts) on which they're generated. As an open-source project, we don't want to, and can't, dictate the platform used by contributors. (I use Ubuntu. Most devs these days have a Mac. Some devs, and most users of web sites, use Windows. And the real users of our apps are using their phones.)

The suggested solution is to generate the expected results via a GitHub action, so anyone can generate a PR that updates the expected screenshots in a consistent way, regardless of their O/S and browser of choice. The goal of this issue is to create or update GitHub actions triggered on creating or updating a PR, that will run the system tests, and if the screenshots change, add the changed screenshot to the PR. 

It will then be the responsibility of the reviewers of the PR to ensure that the changes are acceptable. Changes to screenshots would be acceptable if the PR is actually fixing a bug, where the previous screenshot was actually unacceptable. Changes to screenshots should usually be taken as breaking changes, and we should question ourselves before accepting them. However, in the short term, we can accept more changes as we learn about the testing approach and discover that we need to allow changes to the expected screenshots.

Out of scope for this issue is the question of cross-browser testing. In theory, that's one of the things that Bootstrap should be handling for us. And if there are issues, they're probably at the level of Bootstrap, not our gem. But it's something to think about.

Issue #714, opened 2023-11-04
Some Examples May Be Failing


lcreid:
It appears that examples 50 and 51 in the `README.md` are blank, when they should be showing an error message. Curiously, example 52 works.

Some work was done over the last year that changed how things get rendered. Also, the fact that 52 works with a different class on the div is very suspicious.

PR #715, opened 2023-11-04
Generate Screenshots in CI

lcreid:
Generate screenshots in CI, and add a commit if any screenshots change. Add some documentation that explains how to deal with screenshots changing.

There are a couple of examples (50, 51) that appear to be failing. I've added #714 for them. 

I put the sequence of steps to commit the changes in a Rake task, thinking it would allow us to run the identical steps locally, if needed. Also, it's a bit easier to handle the long commit message in Ruby, at least for me. Does that make sense?

The `git push` when screenshots change triggers another CI run. To avoid that, I added `[skip ci]` to the commit message, but I fear that it would prevent CI runs when, for example, one rebases on the upstream main. The last commit will still have `[skip ci]` and so won't run. So I took it out and we'll simply live with double CI. Okay?

Fixes: #711


lcreid:
@donv This might affect your workflow more than mine. What do you think?

.github/workflows/ruby.yml
```diff
@@ -56,3 +56,8 @@ jobs:
       - name: Run tests
         working-directory: demo
         run: bundle exec rake test:all
+      - name: Commit changed screenshots and other files
+        if: failure()
+        working-directory: demo
+        run: bundle exec rake commit
+
```
donv, line 63: This will automatically commit all changed files in the demo directory after a failure?  That does not seem right.

lcreid: I couldn't think of a better way to communicate the changed screenshots back into a place where the dev could review the screenshots and see if their code needed to be fixed. GitHub's diff for the screenshots is actually pretty slick. I admit that the revert step may be a nuisance, or people may forget it. But we should still catch changes when reviewing the PR.

I could make the workflow more specific, so we _only_ save the screenshot changes when it was the screenshot diff that failed.

I'm open for other suggestions about how to generate and capture the screenshots when we create new ones, or the change is intended.

CONTRIBUTING.md
```diff
@@ -50,6 +50,26 @@ You may find the [demo application](#the-demo-application) useful for developmen
...
 
+When you create or update a pull request, GitHub automatically runs tests that generate the screenshots in the [`README.md`](/README.md). If any of the screenshots change, GitHub will add an additional commit with the updated screenshots.
+
+Normally, the screenshots should _not_ change. If the screenshots changed, please review them _carefully_. If something changed that shouldn't have:
+
+```bash
+git pull # to bring the additional commit to your local branch.
+git revert HEAD # to remove the changes.
+```
+
```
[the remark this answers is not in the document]
lcreid: Really good point! Thanks!

lcreid:
I'll merge this and we can discuss better solutions later.

PR #720, opened 2023-11-28
Remove the require expression for action text in lib/bootstrap_form.rb

### LINKED ISSUE #719: Lazy load ActionText helpers
My application doesn't use ActionText.

Either way, Rails bootstrap_form loads it at:

https://github.com/bootstrap-ruby/bootstrap_form/blob/ed3eda885c2145a364e8affd759f42b03ff6ee7c/lib/bootstrap_form.rb#L1-L9

This require statement creates the `ActionText` module which "tricks" other gems into loading their ActionText support. These gems may do something like:

```rb
if defined?(ActionText)
  # Load ActionText support ...
end
```

This can result in a confusing error message later on when this loaded support eventually fails.

One example of this is RailsAdmin:

https://github.com/railsadminteam/rails_admin/blob/13c90fdd95c3755035b44264dc61e4b65c09ff12/app/assets/javascripts/rails_admin/application.js.erb#L27-L30

I'm wondering if bootstrap_form could somehow use lazy loading to prevent creating the global `ActionText` module until necessary. If possible, this would improve compatibility between these gems when ActionText is not used.

jdufresne:
This require was added in e6f5b09209147143ebdb83032a50e9758cfdc2ab to add ActionText support.

Instead of eagerly loading the module, allow the Rails engine subsystem to load the helpers as necessary.

Eagerly loading the module has the downside of defining the module ActionText even when the engine is not in use. Defining the module like this can "trick" other gems into erroneously loading their ActionText support, which may lead to confusing and out of context error messages.

Fixes #719

### DIFF
--- a/lib/bootstrap_form.rb
+++ b/lib/bootstrap_form.rb
@@ -1,13 +1,3 @@
-# NOTE: The rich_text_area and rich_text_area_tag helpers are defined in a file
-# with a different name and not in the usual autoload-reachable way.
-# The following line is definitely need to make `bootstrap_form` work.
-# rubocop:disable Lint/SuppressedException
-begin
-  require "#{Gem::Specification.find_by_name('actiontext').gem_dir}/app/helpers/action_text/tag_helper"
-rescue Gem::MissingSpecError
-end
-# rubocop:enable Lint/SuppressedException
-
 require "action_view"
 require "action_pack"
 require "bootstrap_form/action_view_extensions/form_helper"


jdufresne:
I would appreciate if something that uses ActionText in their project could test this branch an report what errors they see and when.

lcreid:
The demo app (in the `demo` directory) uses `ActionText` and it seems to run fine. I don't have any other current apps that use `ActionText`, but I'll probably merge this and see if we get complaints from people using the code from `main`.

PR #732, opened 2024-06-05
Fix matrix exclusion

lcreid:


lcreid:
Tests are failing until https://github.com/rails/sprockets-rails/pull/525 is merged and released. Temporarily going to hold sprockets at <3.5.0.

lcreid:
Fixed in #733.

PR #739, opened 2024-10-11
Fix: Add missing form-control-color class for color_field

aydinkazim:
Screenshot 2024-10-11 at 15 01 26


lcreid (approved):
Looks good! Thanks so much for this contribution!

PR #749, opened 2025-04-09
Use import without .css

duleorlovic:
Not sure if older Rails versions work like this, but on Rails 8 and propshaft we need to use `@import` without extension

```
# app/assets/stylesheets/application.scss
@import "bootstrap";
@import "rails_bootstrap_forms";
```

If I follow the current readme and use
```
@import "bootstrap";
@import "rails_bootstrap_forms.css";
```
then there is an error in javascript console
```
GET http://localhost:3000/assets/rails_bootstrap_forms.css net::ERR_ABORTED 404 (Not Found)
```


lcreid:
Thanks for this, @duleorlovic . Looking at that whole section, I think we should re-arrange it to be modern Rails first. GitHub doesn't seem to let me make the suggestions on your PR. Do you mind if I put up a new PR that takes your suggestion and brings it first?

I was thinking that we should say right off the top, something like:

> If you're using Propshaft (for example, the Rails 8 default), do it like this:
> ...
> Else if you're using Sprockets (the Rails default before Rails 8), do it like this:

And probably also a note that there isn't really that much in the styles, and they're free to define their own styling for the classes.

duleorlovic:
Sure, it is easier to amend than to create new PR. Let me know if we need to explain more

README.md
```diff
@@ -58,6 +58,15 @@ If you use Rails in the default mode without any pre-processor, you'll have to a
 
 If you followed the [official bootstrap installation guide](https://github.com/twbs/bootstrap-rubygem#a-ruby-on-rails), you'll probably have switched to SCSS. In this case add the following line to your `application.scss`:
 
+
```
lcreid, line 61: I don't have enough GitHub fu to know how to do this better, so what I would suggest is something like move your added lines to after line 52. The current line 53 would become the last paragraph. We'd also have to update the wording, because it's been a long time since that was likely the "default" way of adding.

If you make those changes, then maybe I can make any other small edits via GitHub suggestions.

PR #751, opened 2025-06-06
Allow file to be -pended

walterdavis:
The file input appears to be treated like any of the other input types in Bootstrap 5. This branch simply removes the previous behavior of forcing it to be wrapped in `input_with_error`, which does not allow options, and instead uses `prepend_and_append_input`, which does. This is the same pattern used in the text input, for example. Errors are handled inside of the `prepend_and_append_input` method, so there's no loss of proper feedback after an invalid submit.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,4 +8,4 @@ gem "drb" if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.4.0")
 gem "mutex_m" if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.4.0")
 gem "rails", BootstrapForm::REQUIRED_RAILS_VERSION
 gem "sprockets-rails", require: "sprockets/railtie"
-gem "sqlite3", "~> 1.4"
+gem "sqlite3", ">= 1.4"
--- a/lib/bootstrap_form/inputs/collection_check_boxes.rb
+++ b/lib/bootstrap_form/inputs/collection_check_boxes.rb
@@ -34,11 +34,11 @@ module BootstrapForm
           def field_name_shim(object_name, method_name, *method_names, multiple: false, index: nil)
             names = method_names.map! { |name| "[#{name}]" }.join
             if object_name.blank?
-              "#{method_name}#{names}#{multiple ? '[]' : ''}"
+              "#{method_name}#{names}#{'[]' if multiple}"
             elsif index
-              "#{object_name}[#{index}][#{method_name}]#{names}#{multiple ? '[]' : ''}"
+              "#{object_name}[#{index}][#{method_name}]#{names}#{'[]' if multiple}"
             else
-              "#{object_name}[#{method_name}]#{names}#{multiple ? '[]' : ''}"
+              "#{object_name}[#{method_name}]#{names}#{'[]' if multiple}"
             end
           end
         end
--- a/test/bootstrap_form_group_test.rb
+++ b/test/bootstrap_form_group_test.rb
@@ -195,6 +195,32 @@ class BootstrapFormGroupTest < ActionView::TestCase
     assert_equivalent_html expected, bootstrap_form_for(@user) { |f| f.text_field :email, prepend: "$", append: ".00" }
   end
 
+  test "file field with prepend text" do
+    expected = <<~HTML
+      <div class="mb-3">
+        <label class="form-label" for="user_avatar">Avatar</label>
+        <div class="input-group">
+          <span class="input-group-text">before</span>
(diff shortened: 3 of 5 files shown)


walterdavis:
@lcreid Please let me know if you would like any additional tests, or for me to squash the commits etc.

test/bootstrap_form_group_test.rb
```diff
@@ -195,6 +195,51 @@ class BootstrapFormGroupTest < ActionView::TestCase
...
+    button_append = button_src
+    suffix = "</div></div>"
+    after_button = prefix + field + button_append + suffix
+    before_button = prefix + button_prepend + field + suffix
+    both_button = prefix + button_prepend + field + button_append + suffix
+    multiple_button = prefix + button_prepend + button_prepend + field + button_append + button_append + suffix
+    assert_equivalent_html after_button, @builder.file_field(:avatar, append: button_src)
+    assert_equivalent_html before_button, @builder.file_field(:avatar, prepend: button_src)
+    assert_equivalent_html both_button, @builder.file_field(:avatar, append: button_src, prepend: button_src)
+    assert_equivalent_html multiple_button, @builder.file_field(:avatar, append: [button_src] * 2, prepend: [button_src] * 2)
```
lcreid, line 240: Nice set of tests!

Gemfile
```diff
@@ -8,4 +8,4 @@ gem "drb" if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.4.0")
 gem "mutex_m" if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.4.0")
 gem "rails", BootstrapForm::REQUIRED_RAILS_VERSION
 gem "sprockets-rails", require: "sprockets/railtie"
-gem "sqlite3", "~> 1.4"
+gem "sqlite3", ">= 1.4"
```
lcreid, line 11: Did you find you needed to change this? There was some dependency issue for Rails 7.1 and earlier where we have to keep SQLite back at 1.4, but if it caused you problems, leave it in. If it didn't actually break anything, let's leave it the way it was, and we'll take it out when we drop support for Rails 7.1.

lcreid (commented):
Looks great. Thank you. One comment that you can respond to according to your judgement, and then we'll merge.

Gemfile
walterdavis: I did, but possibly just because I am unsure how to route around this problem in my local environment. I kept getting the "you have already activated sqlite [much higher version]" when I ran the tests, both with and without bundle exec. I can take it back out and just not run the tests here...

lib/bootstrap_form/inputs/collection_check_boxes.rb
```diff
@@ -34,11 +34,11 @@ def field_name(method, *methods, multiple: false, index: @options[:index])
...
             names = method_names.map! { |name| "[#{name}]" }.join
             if object_name.blank?
-              "#{method_name}#{names}#{multiple ? '[]' : ''}"
+              "#{method_name}#{names}#{'[]' if multiple}"
             elsif index
-              "#{object_name}[#{index}][#{method_name}]#{names}#{multiple ? '[]' : ''}"
+              "#{object_name}[#{index}][#{method_name}]#{names}#{'[]' if multiple}"
             else
-              "#{object_name}[#{method_name}]#{names}#{multiple ? '[]' : ''}"
+              "#{object_name}[#{method_name}]#{names}#{'[]' if multiple}"
```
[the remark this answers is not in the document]
walterdavis: You might think about turning off the auto-correct, then. That would show the findings, and make it more deliberate to fix them in a given patch.

test/bootstrap_form_group_test.rb
walterdavis: Shamelessly duped from a few lines up in the same file! 😁

Gemfile
lcreid: Thanks for the reminder. That was it. Something else already activated a later version. But for some reason we wanted to keep earlier versions of Rails on 1.4. Let's go with what you have and then I'll fix this when we drop Rails 7.1.

lib/bootstrap_form/inputs/collection_check_boxes.rb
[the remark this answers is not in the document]
lcreid: Good suggestion. I'll look into that. I'm not even sure why auto-correct is the default, now that you mention it.

lcreid (approved):
Thanks again for this. I can't promise I'll do a release soon, but you can set your `Gemfile` to pull from `main`. The code doesn't change that much, so it should be low-risk.

PR #764, opened 2025-10-04
Properly handle boolean options in the `fields_for` helper

pienkowb:
This PR fixes how boolean options (e.g. `inline_errors`) are handled in the `fields_for` helper. 

Currently, option defaults are loaded in the helper using the following code:

```ruby
%i[layout control_col inline_errors label_errors].each do |option|
  field_options[option] ||= options[option]
end
```
This approach doesn't work for boolean options, because `false` evalues as falsy, even though it's a legitimate value. As a result, the option value is being overriden with a default value.

For example, the code below does not work as indended (inline errors are still added to the City field):

```ruby
bootstrap_form_for @user, inline_errors: true do |f|
  f.fields_for :address, inline_errors: false do |af|
    af.text_field :city
  end
end
```

In order to fix this, I changed:
```ruby
field_options[option] ||= options[option]
```

to:

```ruby
field_options[option] = field_options.key?(option) ? field_options[option] : options[option]
```

which is more verbose, but works corrently for boolean options.

### DIFF
--- a/demo/app/models/model_user.rb
+++ b/demo/app/models/model_user.rb
@@ -1,5 +1,6 @@
 class ModelUser
   include ActiveModel::Model
+
   attr_accessor :email, :password, :comments, :misc
 
   validates :email, :password, presence: true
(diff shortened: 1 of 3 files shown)


lcreid (commented):
Excellent catch! Thanks for the PR!

PR #779, opened 2025-11-24
Make Collection Controls Aria Accessible

### LINKED ISSUE #763: collection_radio_buttons not accessible
Using the collection_radio_buttons creates a label for the collection as well as a label for each radio button as seen here

https://github.com/bootstrap-ruby/bootstrap_form/blob/f947b147634eb8e5e73c509fc71f165ac270b063/README.md?plain=1#L768-L778
which confuses screenreaders as well as tab navigation. Using a different element like `legend` or allowing some end customization or removal of the label would probably fix. 

[Edit] The same issue exists for checkboxes, so this issue should apply to `collection_checkboxes` as well.

lcreid:
Make `collection_checkboxes` and `collection_radio_buttons` produce output that's more compliant with https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria. 

Since the markup changes noticeably, this change is only enabled by setting the configuration option `group_around_collections = true`:

```
# config/initializers/bootstrap_form.rb

BootstrapForm.configure do |config|
  config.group_around_collections = true
end
```

This PR includes changes in the test files:

- Moved the `collection_checkboxes` and `collection_radio_button` tests to their own files.
- Test `collection_checkboxes` and `collection_radio_button` with the `group_around_collections` options both `true` and `false`. 

Eventually, this behaviour should become the default, and later, the old behavour should disappear altogether.

Closes: #763.


johrstrom:
Hi just want to comment here that I think this is a step in the right direction and to thank you for all the work. Labels for radio buttons are being read now and I believe they fit the WCAG guidelines, though I'm not an expert on the matter.

pablobm:
Why was this made conditional on `Rails.env.development?`? I just arrived here while trying to figure out why my form worked differently in production and in development modes 😅

lcreid:
> Why was this made conditional on `Rails.env.development?`? I just arrived here while trying to figure out why my form worked differently in production and in development modes 😅

First, thanks for your patience while I found some time to reply. I have no idea why I did this, and I normally hate it when I see software that behaves differently in production than it does in development. Let me dig some more and see if I had a reason. Otherwise, I'll fix it. 

Thanks for catching this, and my sincere apologies for the confusion!

lcreid:
@pablobm https://github.com/bootstrap-ruby/bootstrap_form/pull/794. I haven't released this version yet, but you can get the gem from GitHub.

```
gem "bootstrap_form", git: "https://github.com/bootstrap_ruby/bootstrap_form", branch: "main"
```

PR #781, opened 2025-11-27
Adding a configuration option to use turbo_submits_with

pardeyke:
Add custom content to a submit button while the request is processing.
Try `c.turbo_submits_with = :spinner` to check out the build-in spinner or pass a custom html string.

Caveat: Only works on `form.primary` or `form.button` as `form.submit` renders a `input` file which only accepts a string and not a html element. Settings this also overrides `render_as_button` on `form.primary`

Screenshot 2025-11-27 at 01 17 11


lcreid (commented):
Thanks for the PR! This is a very interesting idea. 

I think I'd like to keep this out of a configuration option for now. I'm not saying we won't ever use the configuration options to set "defaults", but I think we should be careful that we don't put everything into the configuration options.

It feels to me like, at least for now, we should try to do this with options to the `bootstrap_form_*` methods, or maybe even just options on the `button` and `primary` methods. Along with a helper method to generate the spinner HTML. Don't forget that not everyone will want the same spinner, and that any text is probably going to depend on the end-developer, not the gem. The most obvious example of what I mean is that you would have to allow any text to be localized to other languages.

What do you think? Would you like to try to narrow the scope of your PR to leave the configuration option for a later PR? Let me know if anything was unclear.

Also, don't forget test cases the next time.

Thanks again for this PR. I think you're on to something.

PR #783, opened 2025-11-27
add option submits_with to button

pardeyke:
as discussed here #781

@lcreid what do you think?

Also should there be a screenshot in den README?

Screenshot 2025-11-27 at 11 01 09


lib/bootstrap_form/inputs/submit.rb
```diff
@@ -4,6 +4,10 @@ module BootstrapForm
   module Inputs
     module Submit
       def button(value=nil, options={}, &)
+        if options.key? :submits_with
+          options[:data] = { turbo_submits_with: setup_turbo_submit(options[:submits_with]) }
```
lcreid: I think this should `merge` with `options[;data]`, in case the user needs to specify other `data-` attributes.

test/bootstrap_other_components_test.rb
```diff
@@ -152,6 +152,22 @@ class BootstrapOtherComponentsTest < ActionView::TestCase
                            @builder.button("<span>I'm HTML!</span> in a button!".html_safe, extra_class: "test-button")
   end
 
+  test "regular button has turbo-submits-with deafault spinner" do
+    expected = <<~HTML
+      <button class="btn btn-secondary" data-turbo-submits-with="<div class=&quot;spinner-border d-block mx-auto&quot; role=&quot;status&quot; style=&quot;--bs-spinner-width: 1lh; --bs-spinner-height: 1lh;&quot;></div>" name="button" type="submit">Submit with Spinner</button>
+    HTML
```
lcreid: I don't think the expected HTML should be escaped.

README.md
```diff
@@ -1065,6 +1065,20 @@ will be rendered as
...
+There is a turbo attribute [(data-turbo-submits-with)](https://turbo.hotwired.dev/reference/attributes) that replaces the content of a submit button while the request is processing.
+```erb
+<%= f.button "Save", submits_with: :spinner %>
+```
+
+```html
+<button class="btn btn-secondary" data-turbo-submits-with="<div class="spinner-border d-block mx-auto"; role="status" style="--bs-spinner-width: 1lh; --bs-spinner-height: 1lh;"></div>" name="button" type="submit">Save</button>
+```
+
+Use `submits_with: :spinner` to render a default bootstrap spinner or pass your own HTML.
```
lcreid: It would be nice to show an example where the user provides the HTML.

README.md
```diff
@@ -1065,6 +1065,20 @@ will be rendered as
...
+This only works on `f.button` or `f.primary` not on `f.submit` and forces `render_as_button: true` on `f.primary`.
```
lcreid: I would move this line up to near the top of this section, because some of us don't read the whole section before we start coding. ;-)

lcreid (commented):
Thanks for this PR. This is a lot like what I was describing in your previous PR.

I really like the idea of making it easier to put up a spinner. I will confess that I hadn't even considered how to do this (I don't do enough front-end work these days). At the same time, I think we should think a lot about how to add new features, so we don't end up committing ourselves to supporting something that turns out not to be what most users need.

With that in mind, I wonder what you about trying to reduce the scope of this PR a bit? What I'm thinking is that instead of a new parameter for `f.button` or `f.primary`, we just provide a helper that only generates this HTML for a spinner, and let the user code the `data-turbo-submits-with` attribute themself. Something like:

```
f.button("Save", data: { turbo_submits_with: f.spinner })
```

This would allow us in the future to make the default spinner more configurable (e.g. with keyword parameters with defaults), without breaking existing applications that were happy with the default spinner.

What do you think?

Also, you're right that we should have an example and screenshot for the documentation. Please add that if you feel you a can figure out how from the existing examples. If run into trouble, feel free to ask me for help.

pardeyke:
I like your idea to let the user code the data attribute themselves.

I can't figure out where the f.spinner should live. Can you point me to file and how to set it up?

lcreid:
> I can't figure out where the f.spinner should live. Can you point me to file and how to set it up?

Good question. I suggest putting it in `lib/bootstrap_form/helpers/bootstrap.rb`. There are some methods in that file that are meant to be used by developers using `bootstrap_form`, like `f.spinner`.

pardeyke:
Sorry it took some time... December is always a mess

pardeyke:
@lcreid What do you think? Are we ready to merge? 😃

karaaslanz:
One behavior change here looks broader than the helper-only direction discussed above.
Turbo currently supports data-turbo-submits-with on both <button> and <input type="submit">: its FormSubmission#setSubmitsWith assigns innerHTML for a button and value for an input. So plain-text data-turbo-submits-with already works with the existing submit/input path.
That makes the new options.dig(:data, :turbo_submits_with) branch in primary potentially surprising: current primary renders an <input name="commit" ...>, while the button path renders <button name="button" ...>. Automatically switching element type for any turbo_submits_with value can therefore change both DOM and submitted parameter semantics, even when the value is just text such as "Saving...".
I think the narrower API would be safer: keep f.spinner as the helper, leave primary's existing element-selection behavior alone, and document that HTML spinner content needs f.button / render_as_button: true, while plain text also works with the normal input submitter. A focused regression could assert that a plain-text turbo_submits_with on primary still renders the input/name="commit" path.

PR #794, opened 2026-04-22
Remove conditional on environment

### LINKED ISSUE #792: Remove Conditional on Environment for Setting Configuration Option
As pointed out: https://github.com/bootstrap-ruby/bootstrap_form/pull/779#issuecomment-4243420998, we shouldn't have code that behaves differently in development and production (unless it has to). I have no memory why the original pull request did it, so I'm taking it out here.

lcreid:
#779 introduced a configuration option, that for some reason defaulted differently in development than in production. This was not good. This PR makes the configuration option the same across environments.

Closes: #792

### DIFF
--- a/lib/bootstrap_form/engine.rb
+++ b/lib/bootstrap_form/engine.rb
@@ -9,7 +9,7 @@ module BootstrapForm
 
     config.bootstrap_form = BootstrapForm.config
     config.bootstrap_form.default_form_attributes ||= {}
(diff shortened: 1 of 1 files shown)


pablobm:
Thank you @lcreid! 🙌 Is it still the case that (as mentioned at https://github.com/bootstrap-ruby/bootstrap_form/pull/779) the plan is for this option to default to `true` in the future?

lcreid:
Yes, it makes sense to me that we should be more compliant with ARIA by default, but we're sensitive to not breaking existing users' pages.

PR #796, opened 2026-06-29
fix: make alert message programmatically focusable

G-Rath:
This makes it possible for downstream consumers to use JavaScript to focus on the alert to satisfy https://www.w3.org/WAI/WCAG21/Understanding/on-focus.html

### DIFF
--- a/lib/bootstrap_form/helpers/bootstrap.rb
+++ b/lib/bootstrap_form/helpers/bootstrap.rb
@@ -9,7 +9,7 @@ module BootstrapForm
         css = options[:class] || "alert alert-danger"
         return unless object.respond_to?(:errors) && object.errors.full_messages.any?
 
-        tag.div class: css do
+        tag.div class: css, tabindex: "-1" do
           if options[:error_summary] == false
             title
           else


karaaslanz:
Small review note: this changes the rendered alert_message markup, but the existing alert_message assertions in test/bootstrap_form_test.rb still expect the <div> without tabindex. assert_equivalent_html compares the full Nokogiri fragments and its only loose exception is data-disable-with, so the current expectations should be updated and/or a focused regression should explicitly assert tabindex="-1".
That would both protect the intended accessibility behavior and make the change self-documenting in the test suite.

lcreid:
@G-Rath I'm afraid I haven't dug into accesibility enough to understand why your fix fixes the problem. In fact, I don't understand how the `div` is even taking focus, as it doesn't have a control in it. Or am I missing something?

G-Rath:
Focus is important for those using assistive technology such as screen readers and keyboard-only navigators, as it relates to changes on a page.

For example, consider submitting a form with a validation error - that would take you back to the form with an alert message summarizing the errors. To a non-sighted person if there's no change-in-focus they'll be told a generic "you're on <page you were already on>", and they'll have to hunt around for the alert message.

The ideal is when the page loads the focus is put on the form error summary, so the non-sighted person is told "focus is on: <content of alert>".

Doing this requires making the alert focused, and to do that it has to be _focusable_, and you've said by default `div` elements cannot be focused at all - which is what `tabindex: "-1"` enables.

Note this does not put the element in the tab order, it just means that you can programmatically call `focus` on it.

As a more complete example, I'm working with the NZ govt on a "SaaS" version of [the CWAC accessibility checker](https://github.com/GOVTNZ/cwac); we have a snippet of JS that looks for bootstrap alerts and calls `focus` on the first one on the page.

You can see this in action on our [sign in page](https://cwac.govt.nz/sign_in) (just use any username + password, its a closed beta thing right now), where the "that email or password is incorrect" alert will get focused when the page is loaded, and once it loses focus you cannot tab it back into focus.

G-Rath:
Alternatively (and tbh probably the better solution here) I could add generic support for e.g. `wrapper`, to allow you to apply any attribute to the element.

That would actually us remove our patch completely because we're also needing to set `role="note"` on the message, but I didn't do that here as my understanding its that is far more situational whereas this change should always be appropriate given this is an "alert message"

